PR #5, opened 2019-09-15
solution

youvovk:


src/components/Carousel.js
```diff
@@ -1,17 +1,174 @@
...
+        itemWidth,
+      },
+    } = this;
+
+    const leftButtonStyle = {
+      width: `${itemWidth / 4}px`,
+      height: `${itemWidth / 4}px`,
+      left: `-${itemWidth / 4}px`,
+    };
+    const rightButtonStyle = {
```
irenhh: the variable above is the same, combine them

src/components/Carousel.js
```diff
@@ -1,17 +1,174 @@
...
+          >
+            {visibleImages
+              .map((image, index) => (
+                <option key={image} value={index + 1}>{index + 1}</option>
+              ))
+            }
+          </select>
+
+          <p className="setting-image-width__name">Pictures size:</p>
+          <p>0</p>
```
irenhh: is the first value 0?

src/components/Carousel.js
```diff
@@ -1,17 +1,174 @@
...
+      ? (this.handledScroll.scrollLeft = 0)
+      : (this.handledScroll.scrollLeft
+          += this.state.selectedStep * this.state.itemWidth)
+  );
+
+  getPreviousImages = () => (
+    this.setState(prevState => ({
+      indexFirstImage: this.handledScroll.scrollLeft / prevState.itemWidth,
+    })),
+    this.handledScroll.scrollLeft === 0
```
irenhh: if you want to assign something to a variable with ternary operator, do like in the example (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators#Conditional)
otherwise use if statement

src/components/Carousel.js
```diff
@@ -1,17 +1,174 @@
...
+          />
+          <div
+            className="Carousel__list-wrapper"
+            style={{ width: `${selectedFrameSize * itemWidth}px` }}
+            ref={(div) => { this.handledScroll = div; }}
+          >
+
+            <ul id="container" className="Carousel__list">
+              {
+                visibleImages.map((image, index) => (
```
irenhh: no need to place it on the next line, move it to the opened bracket

PR #53, opened 2021-01-15
testing commit

Lara-l-art:
[DEMO LINK](https://Lara-l-art.github.io/react_carousel/)

It isn't a perfect solution, now I guess I can improve this and create this pull request for check table synchronization

### DIFF
--- a/src/components/Carousel.js
+++ b/src/components/Carousel.js
@@ -1,19 +1,48 @@
 import React from 'react';
-
+import PropTypes from 'prop-types';
 import './Carousel.scss';
 
-const Carousel = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
-      <li><img src="./img/1.png" alt="3" /></li>
-      <li><img src="./img/1.png" alt="4" /></li>
-    </ul>
+const Carousel = ({ images, width, count }) => (
+  <>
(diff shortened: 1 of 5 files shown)


tamara-ostapets (approved):
great work 🌟

vsm2k17 (changes requested):
good job but let's make it better

src/components/Carousel.js
```diff
@@ -1,19 +1,122 @@
...
-    </ul>
+class Carousel extends React.Component {
+  state = {
+    itemWidth: this.props.itemWidth,
+    frameSize: this.props.frameSize,
+    step: this.props.step,
+    position: 0,
+    animationDuration: this.props.animationDuration,
+    infinite: this.props.infinite,
+    maxPosition: this.props.itemWidth * this.props.images.length
```
vsm2k17, line 13: you don't need to save props inside of a state, that's a bad idea use props as an initial state, and a redundant step because you have access to your props anyway

vsm2k17 (changes requested):
good job, but it would be great if a user will be able to change those metrics, so let's add form and change props for the carousel component when the user changes them in form.

vsm2k17 (changes requested):
Uploading Screen Recording 2021-01-21 at 20.35.00.mov…

great work,
I found some bugs, fix them

vsm2k17 (commented):
https://user-images.githubusercontent.com/35385361/105396574-be888f80-5c28-11eb-9c89-3e6a1546650f.mov

PR #55, opened 2021-01-18
Solution dirty

AlexysUa:
[DEMO LINK](https://alexysua.github.io/react_carousel/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -27,7 +27,14 @@ class App extends React.Component {
         {/* eslint-disable-next-line */}
         <h1>Carousel with {images.length} images</h1>
 
-        <Carousel />
+        <Carousel
+          images={images}
+          step={4}
+          frameSize={4}
+          itemWidth={130}
+          animationDuration={1000}
+          infinite={false}
+        />
       </div>
     );
   }
--- a/src/components/Carousel.css
+++ b/src/components/Carousel.css
@@ -1,0 +1,6 @@
+.Carousel__list {
+  display: flex;
+  list-style: none;
+  padding-inline-start: 0;
+  overflow: hidden;
+}
--- a/src/components/Carousel.js
+++ b/src/components/Carousel.js
@@ -1,17 +1,119 @@
 import React from 'react';
+import './Carousel.css';
+import PropTypes from 'prop-types';
 
-const Carousel = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
-      <li><img src="./img/1.png" alt="3" /></li>
-      <li><img src="./img/1.png" alt="4" /></li>
-    </ul>
+class Carousel extends React.Component {
+  state = {
+    imagesWidth: 1300,
+    images: [...this.props.images],
+    step: this.props.step,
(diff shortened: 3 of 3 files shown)


tamara-ostapets (approved):
cool work 👀

You could do props validation as a static property of class-component (idk if it's a better way, but convenient). Interesting, what mentors have to say about it.
```
class Carousel extends React.Component {
  static propTypes {
    images: PropTypes.arrayOf(PropTypes.string).isRequired,
    ...
    }
    ...
}
```

AlexysUa:
> cool work 👀
> 
> You could do props validation as a static property of class-component (idk if it's a better way, but convenient). Interesting, what mentors have to say about it.
> 
> ```
> class Carousel extends React.Component {
>   static propTypes {
>     images: PropTypes.arrayOf(PropTypes.string).isRequired,
>     ...
>     }
>     ...
> }
> ```

Thanks for the review! First of all, I want to visit QA and ask about passing props to the state.

vsm2k17:
> > cool work 👀
> > You could do props validation as a static property of class-component (idk if it's a better way, but convenient). Interesting, what mentors have to say about it.
> > ```
> > class Carousel extends React.Component {
> >   static propTypes {
> >     images: PropTypes.arrayOf(PropTypes.string).isRequired,
> >     ...
> >     }
> >     ...
> > }
> > ```
> 
> Thanks for the review! First of all, I want to visit QA and ask about passing props to the state.

bad idea don't use props as an initial state

AlexysUa:
> > > cool work 👀
> > > You could do props validation as a static property of class-component (idk if it's a better way, but convenient). Interesting, what mentors have to say about it.
> > > ```
> > > class Carousel extends React.Component {
> > >   static propTypes {
> > >     images: PropTypes.arrayOf(PropTypes.string).isRequired,
> > >     ...
> > >     }
> > >     ...
> > > }
> > > ```
> > 
> > 
> > Thanks for the review! First of all, I want to visit QA and ask about passing props to the state.
> 
> bad idea don't use props as an initial state

I will fix it! Should I add input form for users?

src/App.js
```diff
@@ -17,17 +18,101 @@ class App extends React.Component {
...
+    frameSize: 3,
+    itemWidth: 130,
+    animationDuration: 1000,
+    infinite: false,
+    counter: 3,
+    translateX: 0,
   };
 
+  paramFromInputs
+    = (frameSize, itemWidth, animationDuration, step, infinite, counter) => {
```
vsm2k17, line 31:
```suggestion
  paramFromInputs = (
    frameSize, 
    itemWidth, 
    animationDuration, 
    step, 
    infinite, 
    counter,
  ) => {
```

src/components/Form.js
```diff
@@ -0,0 +1,120 @@
...
+      step,
+      infinite,
+      counter,
+    );
+  }
+
+  handleChange = (event) => {
+    const { name, value, type, checked } = event.target;
+
+    // eslint-disable-next-line no-nested-ternary
```
vsm2k17, line 40: no need to disable eslint, create a switch construction better

vsm2k17 (approved):
image
Validation of form would be a great idea, but that's not necessary.
The infinity mode of your carousel very good but infinity means that you can press the left button and scrolling items infinity times with no returning back, it's hard to implement, but your version is also great.

Looks great 😎

PR #57, opened 2021-01-19
Solution

sasha39612:
[DEMO LINK](https://sasha39612.github.io/react_carousel/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import './App.scss';
-
 import Carousel from './components/Carousel';
 
 class App extends React.Component {
@@ -25,9 +24,18 @@ class App extends React.Component {
     return (
       <div className="App">
         {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1 className="main-title">
+          {`Carousel with ${images.length} images`}
+        </h1>
 
-        <Carousel />
+        <Carousel
+          images={images}
+          step={3}
+          frameSize={4}
+          itemWidth={130}
+          animationDuration={1000}
+          infinite={false}
+        />
       </div>
     );
   }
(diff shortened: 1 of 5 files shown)


src/components/Carousel.js
```diff
@@ -1,19 +1,121 @@
...
+
+          return { translation: startPosition };
+        }
+      }
+
+      this.prevPosition = position;
+
+      return { translation: position };
+    });
+  }
```
varseniuk, line 36: You are creating two almost similar functions. To avoid this you could create only one function and name it 'scroll', for example. This function will take one argument (string 'forward' or 'backward') and with 'switch...case' construction you could define what to do.
E.g.

```
scroll = (direction) => {
  const { step, images, itemWidth, frameSize, infinite } = this.props;

  switch (direction) {
    case 'forward':
   ....some code...
    case 'backward':
   ....some code...
  }

  this.prevPosition = position;
  return { translation: position };
}
```

and call this function on your 'onClick' handler:
`onClick={() => this.scroll('forward')}`
or
`onClick={() => this.scroll('backward')}`

src/App.js
```diff
@@ -25,9 +24,18 @@ class App extends React.Component {
...
+        </h1>
 
-        <Carousel />
+        <Carousel
+          images={images}
+          step={3}
+          frameSize={4}
+          itemWidth={130}
+          animationDuration={1000}
+          infinite={false}
```
varseniuk, line 37: You could create another component carousel with 'infinite = {true}' so we could see the difference.
Your code is working, I've tried, but it would be better if you add another carousel to show the difference

varseniuk (approved):
Good job! 👍 
But you could make it even better. Read my comments please

sasha39612:
Vitalii thank you for good advices!

PR #82, opened 2021-05-25
develop

Andrey203:
[DEMO LINK](https://Andrey203.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.js
+++ b/src/components/Carousel.js
@@ -1,19 +1,130 @@
 import React from 'react';
+import PropTypes, { string } from 'prop-types';
 
 import './Carousel.scss';
 
-const Carousel = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
-      <li><img src="./img/1.png" alt="3" /></li>
-      <li><img src="./img/1.png" alt="4" /></li>
-    </ul>
+class Carousel extends React.Component {
+  state = {
+    direction: 'right',
+    left: 0,
+  }
 
-    <button type="button">Prev</button>
-    <button type="button">Next</button>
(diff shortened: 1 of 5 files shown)


SerhiiKolesnyk (approved):
круто_)

src/components/Carousel.js
```diff
@@ -1,19 +1,130 @@
...
+  frameWidth = this.props.frameSize * this.props.itemWidth;
+
+  step = this.props.step * this.props.itemWidth;
+
+  maxTransition = this.props.imgList.length * this.props.itemWidth
+    - this.frameWidth;
+
+  componentDidMount() {
+    this.carousel = document.querySelector('.Carousel__list');
+    this.carouselWrapper = document.querySelector('.Carousel__wrapper');
```
VinnieJ-2k20, line 27: Let's not work directly with DOM in React.
You can form a style object and then apply it to element directly:

```
<div
  style={styleObj}
>
  ...
</div>
```

src/components/Carousel.js
```diff
@@ -1,19 +1,130 @@
...
+  }
+}
+
+Carousel.propTypes = {
+  imgList: PropTypes.arrayOf(string).isRequired,
+  step: PropTypes.number.isRequired,
+  frameSize: PropTypes.number.isRequired,
+  itemWidth: PropTypes.number.isRequired,
+  animationDuration: PropTypes.number.isRequired,
+  infinite: PropTypes.bool.isRequired,
```
VinnieJ-2k20, line 127: You have default values for `itemWidth`, `frameSize`, `step`, `animationDuration` and `infinite` props.
Therefore, these fields are not required.

VinnieJ-2k20 (changes requested):
Nice working solution.
But you need to change styling approach and avoid working with DOM directly.
Also, it would be great to disable buttons if move in the corresponding direction is not possible.

PR #107, opened 2021-12-08
Solution without infinity loop

Vitaliy-0:
[demo](https://vitaliy-0.github.io/react_carousel/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,5 @@
-// style go here
+.App {
+  &__title {
+    text-align: center;
+  }
+}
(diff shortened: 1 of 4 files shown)


ihor-jpeg (changes requested):
add an ability to customize itemWidth with default value of 130px
add frameSize - number of images displayed at the same time with the default of 3
add step (default 3) - number of images scrolled per click
add animationDuration (default 1000) - time in ms to show the new portion of images

AlinaMatuschak (approved):

PR #108, opened 2021-12-11
Carousel

dariia-romanova:
[DEMO LINK](https://dariia-romanova.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,74 @@
 import React from 'react';
+import { CarouselType } from '../types/CarouselType';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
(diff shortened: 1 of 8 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,74 @@
...
+      if (carousel.current.scrollLeft !== 0) {
+        carousel.current.scrollLeft -= step * itemWidth
+          + (carousel.current.scrollLeft % itemWidth);
+      } else {
+        carousel.current.scrollLeft = (images.length - frameSize) * itemWidth;
+      }
+    }
+  };
+
+  React.useEffect(() => {
```
dmtrhrytsak, line 34: Better import hooks at the top of your program, don't access it directly

dmtrhrytsak (approved):
It scrolls and that's all that matters. Nice!

ihor-jpeg (changes requested):
add an ability to customize itemWidth with default value of 130px
add frameSize - number of images displayed at the same time with the default of 3
add step (default 3) - number of images scrolled per click
add animationDuration (default 1000) - time in ms to show the new portion of images

PR #118, opened 2022-02-07
Carousel

InnaPogrebna:
DEMO LINK
https://InnaPogrebna.github.io/react_carousel/

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,232 @@
 import React from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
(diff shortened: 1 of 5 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,232 @@
...
+      this.setState({
+        disabledPrev: true,
+      });
+    }
+  };
+
+  getChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { name } = event.target;
+
+    this.setState({ [name]: +event.target.value } as unknown as Pick<State, keyof State>);
```
danylo-onishchuk, line 85: why as unknown?

danylo-onishchuk (approved):
great job at all!

fix ts issue

src/components/Carousel.tsx
InnaPogrebna: Honestly, I used google to solve this problem, because I had a mistake.  When I used "as unknown" this mistake disappeared. I know that I need to use something different but I don't have any idea.  Could you tell me what I need to use instead of it?

PR #174, opened 2022-07-04
React carousel

vsupruniuk:
[DEMO LINK](https://vsupruniuk.github.io/react_carousel/)

Infinite scrolling isn't perfect, but it's the best I can do☹

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,32 @@
-// style go here
+.App {
+  margin: 0 auto;
+
+  &__title {
+    text-align: center;
+    margin-bottom: 50px;
+  }
+}
+
+.control {
+  margin: 0 auto;
+  width: 350px;
+
+  &__list {
+    display: flex;
+    flex-direction: column;
+    list-style: none;
+    margin: 0;
+  }
+
+  &__item {
+    display: flex;
+    justify-content: space-between;
+    font-size: 20px;
+    margin-bottom: 15px;
+  }
+
+  &__input {
+    width: 100px;
+    text-align: center;
+  }
+}
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,4 +1,4 @@
-.App {
+.app {
   margin: 0 auto;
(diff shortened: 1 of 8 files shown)


src/App.scss
```diff
@@ -1 +1,32 @@
-// style go here
+.App {
+  margin: 0 auto;
```
jstmpelowycz, line 2: BEM?
image

jstmpelowycz: the same way all over your code, make sure you fix it

src/App.tsx
```diff
@@ -20,17 +26,132 @@ class App extends React.Component<{}, State> {
...
+      step,
+      animationDuration,
+      isInfinite,
+    } = this.state;
 
     return (
       <div className="App">
         {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1 className="App__title">Carousel with {images.length} images</h1>
```
jstmpelowycz, line 69:
```suggestion
        <h1 className="App__title">
          {`Carousel with ${images.length} images`}
        </h1>
```
I suppose `eslint-disable` is because of your approach of string output

src/App.tsx
```diff
@@ -20,17 +26,132 @@ class App extends React.Component<{}, State> {
...
+              <input
+                className="control__input"
+                type="number"
+                value={itemWidth}
+                min="100"
+                max="200"
+                step="10"
+                onChange={({ target }) => {
+                  this.changeItemWidth(Number(target.value));
+                }}
```
jstmpelowycz, line 93: you could create a separate callback, and just its ref here, it would look more common
```
const handleItemWidthChange = () => {
  ...
}
```

jstmpelowycz: and the same way all over your code

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #176, opened 2022-07-10
Solution

stanislav-topikha:
[DEMO LINK](https://stanislav-topikha.github.io/react_carousel/)

image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,7 +3,14 @@ import './App.scss';
 import Carousel from './components/Carousel';
 
 interface State {
-  images: string[];
+  images: string[],
+  userNumInputs: {
+    step: number,
+    frameSize: number,
+    itemWidth: number,
+    animationDuration: number,
+  },
+  infinite: boolean,
 }
 
 class App extends React.Component<{}, State> {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -20,17 +27,91 @@ class App extends React.Component<{}, State> {
...
+    userNumInputs: {
+      step: 3,
+      frameSize: 3,
+      itemWidth: 130,
+      animationDuration: 1000,
+    },
+    infinite: false,
   };
 
+  numInputHandler(value: string, key: string) {
```
Olexandr6, line 39: Read more about naming handler functions - https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364

Fix this for all handlers

src/App.tsx
```diff
@@ -20,17 +27,91 @@ class App extends React.Component<{}, State> {
...
+          animationDuration={userNumInputs.animationDuration}
+          infinite={infinite}
+        />
+
+        <form className="app__form">
+          {(Object.keys(userNumInputs) as (keyof typeof userNumInputs)[])
+            .map(key => (
+              <label key={key} className="app__input">
+                <span>
+                  {`${key.split(/(?=[A-Z])/).join(' ').toUpperCase()} : `}
```
Olexandr6, line 89: It looks too complex, I would better render exact number of inputs without mapping, this mapping makes your solution too complex

Olexandr6 (changes requested):
Woww, looks great🔥 It's not an easy task - well done👍 Let's fix some things👇

src/App.tsx
Olexandr6: Or find way to simplify this

src/App.tsx
stanislav-topikha: fixed

src/App.tsx
stanislav-topikha: simplified

Olexandr6 (approved):
🔥

PR #181, opened 2022-07-28
cant find away for infinite carousel

Havrysh:
Tried to import 'infinite-react-carousel' but couldn't figure out how to do it correctly with typings and module exporting. Maybe there is an easier way to make it work


Havrysh:
https://havrysh.github.io/react_carousel/

ArturErmolenko (approved):
looks good 👍

PR #193, opened 2022-08-09
added carousel

silchyn:
[DEMO](https://silchyn.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -52,28 +52,40 @@ class App extends React.Component<{}, State> {
       return;
     }
 
-    if (target.classList.contains('form__itemWidth')) {
-      this.setState({
-        itemWidth: Math.min(Math.max(+target.value, 80), 200),
-      });
+    switch (true) {
+      case target.classList.contains('form__itemWidth'):
+        this.setState({
+          itemWidth: Math.min(Math.max(+target.value, 80), 200),
+        });
 
-      target.value = this.state.itemWidth.toString();
-    } else if (target.classList.contains('form__frameSize')) {
-      this.setState({
-        frameSize: Math.min(Math.max(+target.value, 1), imagesLength),
-      });
+        target.value = this.state.itemWidth.toString();
 
-      target.value = this.state.frameSize.toString();
-    } else if (target.classList.contains('form__step')) {
-      this.setState({ step: Math.max(+target.value, 1) });
+        break;
+      case target.classList.contains('form__frameSize'):
+        this.setState({
+          frameSize: Math.min(Math.max(+target.value, 1), imagesLength),
+        });
 
-      target.value = this.state.step.toString();
-    } else if (target.classList.contains('form__animationDuration')) {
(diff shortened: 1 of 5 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,114 @@
...
+          className="Carousel__list"
+          style={{
+            transform: `translateX(-${itemWidth * position}px)`,
+            width: itemWidth * images.length,
+            transition: `transform ${animationDuration}ms`,
+          }}
+        >
+          {
+            images
+              .map((image, i) => (
```
ivan-kadykalo, line 78:
```suggestion
            images.map((image, i) => (
```

src/App.tsx
```diff
@@ -20,17 +25,146 @@ class App extends React.Component<{}, State> {
...
+
+        target.value = this.state.step.toString();
+      } else if (target.classList.contains('form__animationDuration')) {
+        this.setState({ animationDuration: Math.max(+target.value, 0) });
+
+        target.value = this.state.animationDuration.toString();
+      } else if (target.classList.contains('form__infinite')) {
+        this.setState({ infinite: target.checked });
+      }
+    }
```
ivan-kadykalo, line 75: 🤯 Avoid nested ifs and use switch case

ivan-kadykalo (changes requested):

src/App.tsx
silchyn: > 🤯 Avoid nested ifs and use switch case

But how to use switch case with .contains()?

ivan-kadykalo: May google be with you🙃

image

src/components/Carousel.tsx
```diff
@@ -1,18 +1,113 @@
...
+    const buttons = document.querySelector('.Carousel__buttons');
+
+    if (buttons) {
+      buttons.children[0].addEventListener('click', () => {
+        this.setState((state, props) => {
+          if (props.infinite && !state.position) {
+            return { position: props.images.length - props.frameSize };
+          }
+
+          return { position: Math.max(0, state.position - props.step) };
```
ivan-kadykalo, line 34:
```suggestion
          return (props.infinite && !state.position) 
          	? { position: props.images.length - props.frameSize }
          	: { position: Math.max(0, state.position - props.step) };
```

ivan-kadykalo (changes requested):
Fix switch case

src/App.tsx
silchyn: > May google be with you🙃
> 
> image

Yeah, I know about that kludge but why it is better than if-else-if?

ivan-kadykalo: Switch case is more readable and works faster than if else.

ivan-kadykalo: It is not a better solution each time, but in this case it is.

Ihor-Karpyn (approved):

ivan-kadykalo (approved):

PR #198, opened 2022-08-18
React Carousel component was created

dmytro-lebedchenko:
[DEMO LINK](https://dmytro-lebedchenko.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -23,41 +23,32 @@ export class Carousel extends React.Component<Props, State> {
     const {
       images,
       step,
-      infinite,
     } = this.props;
 
     const { position } = this.state;
 
-    switch (true) {
-      case (type === 'previous') && (position - step >= 0):
+    if (type === 'previous') {
+      if (position - step >= 0) {
         this.setState(state => ({
           position: state.position - step,
         }));
-        break;
-
-      case (type === 'previous') && infinite:
+      } else {
         this.setState({
           position: (images.length - step),
         });
-        break;
+      }
+    }
 
-      case (type === 'next') && ((position + step) < images.length):
+    if (type === 'next') {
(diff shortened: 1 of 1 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,117 @@
...
+    }
+
+    if (type === 'next') {
+      if ((position + step) < images.length) {
+        this.setState({ position: position + step });
+      } else if (infinite) {
+        this.setState({ position: 0 });
+      }
+    }
+  };
```
Bohdan-Maksymiuk, line 48: avoid such nested `if else` conditions
This code is hard to read and maintain

Bohdan-Maksymiuk (changes requested):
https://user-images.githubusercontent.com/94289038/185951540-6785fb2b-eb70-47a7-b582-074f2ae6943e.mov

here should be 3 icons

dmytro-lebedchenko:
> Screen.Recording.2022-08-22.at.17.51.42.mov 
> here should be 3 icons

Thank you for your recommendations! 
All changes has been implemented ✅

src/components/Carousel.tsx
```diff
@@ -1,18 +1,145 @@
...
+  buttonHandler = (type: string) => {
+    const {
+      images,
+      step,
+      infinite,
+    } = this.props;
+
+    const { position } = this.state;
+
+    switch (true) {
```
SanyaBratashchuk, line 31: Try to rewrite this logic without switch(true) that looks it looks strange)

SanyaBratashchuk (changes requested):

src/components/Carousel.tsx
dmytro-lebedchenko: Done ✅

ArturErmolenko (approved):

PR #200, opened 2022-08-23
Solution

OlhaCh:
[DEMO LINK](https://olhach.github.io/react_carousel/)


src/App.tsx
```diff
@@ -20,17 +26,120 @@ class App extends React.Component<{}, State> {
       './img/9.png',
       './img/10.png',
     ],
+    step: '3',
+    frameSize: '3',
+    itemWidth: '130',
+    animationDuration: '1000',
+    infinite: false,
```
SanyaBratashchuk, line 33: You doesn't use it

src/components/Carousel.tsx
```diff
@@ -1,18 +1,124 @@
...
-      <li><img src="./img/1.png" alt="4" /></li>
-    </ul>
-
-    <button type="button">Prev</button>
-    <button type="button">Next</button>
-  </div>
-);
-
-export default Carousel;
+type Props = {
```
SanyaBratashchuk, line 4: Use interface for Props too

src/components/Carousel.tsx
```diff
@@ -1,18 +1,124 @@
...
+interface State {
+  shift: number,
+  isNextDisabled: boolean,
+}
+
+export class Carousel extends Component<Props, State> {
+  state = {
+    shift: 0,
+    isNextDisabled: false,
+    isPrevDisabled: true,
```
SanyaBratashchuk, line 21: You don't need to hold in the state the things you can calculate using available state

SanyaBratashchuk (changes requested):

PR #204, opened 2022-09-05
solution

tiserett:
[DEMO LINK](https://tiserett.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,27 +3,31 @@ import './App.scss';
 import { Carousel } from './components/Carousel';
 
 interface State {
-  images?: string[];
-  itemWidth?: number;
-  frameSize?: number;
-  step?: number;
-  animationDuration?: number;
+  itemWidth: number;
+  frameSize: number;
+  step: number;
+  animationDuration: number;
 }
 
-export class App extends Component<{}, State> {
+type Props = {
+  images: string[]
+};
+
+export const imagesList = [
+  './img/1.png',
+  './img/2.png',
+  './img/3.png',
+  './img/4.png',
+  './img/5.png',
+  './img/6.png',
+  './img/7.png',
+  './img/8.png',
+  './img/9.png',
+  './img/10.png',
+];
+
+export class App extends Component<Props, State> {
   state = {
-    images: [
-      './img/1.png',
-      './img/2.png',
-      './img/3.png',
-      './img/4.png',
-      './img/5.png',
-      './img/6.png',
-      './img/7.png',
-      './img/8.png',
-      './img/9.png',
-      './img/10.png',
-    ],
     itemWidth: 130,
     frameSize: 3,
     step: 3,
@@ -33,17 +37,23 @@ export class App extends Component<{}, State> {
   handleInput = (event: React.ChangeEvent<HTMLInputElement>) => {
     const { name, value } = event.target;
 
-    this.setState({ [name]: +(value) });
+    this.setState(prevState => (
+      {
+        ...prevState,
+        [name]: +value,
+      }
+    ));
   };
 
   render() {
     const {
-      images,
       itemWidth,
       frameSize,
       step,
       animationDuration,
     } = this.state;
+
+    const { images } = this.props;
 
     return (
       <div className="App">
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -57,8 +57,9 @@ export class App extends Component<Props, State> {
 
     return (
       <div className="App">
-        {/* eslint-disable-next-line */}
-        <h1 title="carousel">Carousel with {images.length} images</h1>
+        <h1 title="carousel">
+          {`Carousel with ${images.length} images`}
+        </h1>
 
         <div className="labels">
           <label className="labels__item">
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,12 +1,16 @@
...
-import Carousel from './components/Carousel';
+import { Carousel } from './components/Carousel';
 
 interface State {
-  images: string[];
+  images?: string[];
+  itemWidth?: number;
+  frameSize?: number;
+  step?: number;
+  animationDuration?: number;
```
sergii-nosachenko, line 10: These properties are not optional. On line 14 you declare all of them as default state values for your component.

src/components/Carousel.scss
```diff
@@ -1 +1,37 @@
...
+  }
+}
+
+.btn-prev:hover {
+  transform: scale(1.2);
+}
+
+.btn-next:hover {
+  transform: scale(1.2);
+}
```
sergii-nosachenko, line 37:
```suggestion
  &__btn {
    display: flex;
    justify-content: center;
    align-items: center;
    padding: 10px 30px;
    border-radius: 20px;

    transition: transform 0.3s linear;
    
    &--prev,
    &--next {
      transform: scale(1.2);
    }
  }
}
```

Your BEM is bad, Maks, I'm judging you 😈
Use modifiers on your buttons.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,130 @@
...
+  canMovePrev: boolean;
+  canMoveNext: boolean;
+};
+
+export class Carousel extends Component<Props, State> {
+  state = {
+    translate: 0,
+    gap: 30,
+    canMovePrev: false,
+    canMoveNext: true,
```
sergii-nosachenko, line 24: These default values don't prevent you against mistakes.
If you have only 3 items in total in 3 pieces frame you'll have next button active.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,130 @@
...
+        translate: -limit,
+        canMovePrev: true,
+        canMoveNext: false,
+      });
+    } else {
+      this.setState({
+        translate: nextStep,
+        canMovePrev: true,
+      });
+    }
```
sergii-nosachenko, line 67: I think you combine it in one setState call.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,130 @@
...
+      canMoveNext,
+    } = this.state;
+
+    const {
+      images,
+      itemWidth,
+      frameSize,
+      animationDuration,
+    } = this.props;
+
```
sergii-nosachenko, line 84: Something wrong is going on inside your handlers. Look [here](https://i.imgur.com/hLZns3S.png) and [here](https://i.imgur.com/xuM6eUb.png). Try to play with options and window width to catch these bugs.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,130 @@
...
+    return (
+      <div className="Carousel">
+        <button
+          className="Carousel__btn btn-prev"
+          type="button"
+          onClick={this.handleClickPrev}
+          disabled={!canMovePrev}
+        >
+          Prev
```
sergii-nosachenko, line 93: It seems that you need to replace texts with arrows due the task condition.

sergii-nosachenko (changes requested):
Maks, I've found some mistakes and bugs 🐛🐞🐜
Sorry, but you need to fix it.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,121 @@
...
+
+    const nextStep = translate - step * itemWidth - gap * step;
+
+    this.setState({
+      translate: (nextStep < -limit) ? -limit : nextStep,
+      canMovePrev: true,
+      canMoveNext: !(nextStep < -limit),
+    });
+  };
+
```
sergii-nosachenko, line 60: Maks, it seems I've found the reason of bugs. Inside your handlers you determine the values from current state and props and then set new values. But! A huge issue here is that React optimizes the process of rendering, so now you don't know what values will be calculated (state can change suddenly during your code execution).
I propose you to use setState with callback immediately after you get inside your handler:
```
handler = () => this.setState((state, props) => {
  // calculations
  
  return {
    // new state object
  }
})
```

Why this is important? You can read it [here](https://uk.reactjs.org/docs/react-component.html#setstate)
Briefly, you need better control over what's going on when you want to set state.

sergii-nosachenko (changes requested):
Code become better, but I found the source of your bugs.

sergii-nosachenko (approved):
Well done 🔥

src/App.tsx
```diff
@@ -1,39 +1,120 @@
...
+      frameSize,
+      step,
+      animationDuration,
+    } = this.state;
+
+    const { images } = this.props;
 
     return (
       <div className="App">
         {/* eslint-disable-next-line */}
```
danylo-onishchuk, line 60: ?

danylo-onishchuk (approved):
good job at all :)

PR #209, opened 2022-09-12
Carousel implemented

andrii-dmytruk:
[DEMO LINK](https://andrii-dmytruk.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -59,10 +59,18 @@ class Carousel extends React.Component<Props, State> {
         leftDisable: false,
       });
     } else {
-      this.setState({
-        translate: infinite ? 0 : maxShift,
-        rightDisable: !infinite,
-      });
+      // eslint-disable-next-line no-lonely-if
+      if (prevTranslate === maxShift && infinite) {
+        this.setState({
+          translate: 0,
+          rightDisable: !infinite,
+        });
+      } else {
+        this.setState({
+          translate: maxShift,
+          rightDisable: !infinite,
+        });
+      }
     }
   };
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -20,17 +25,112 @@ class App extends React.Component<{}, State> {
...
-        <h1>Carousel with {images.length} images</h1>
+        <h1 data-cy="title">Carousel with {images.length} images</h1>
+        <div className="controls">
+          <label>
+            {'ItemWidth: '}
+            <input
+              type="number"
+              name="itemWidth"
+              id="ItemWidth"
+              defaultValue="130"
```
sergii-nosachenko, line 76:
```suggestion
              value={itemWidth}
```
Better use value from state, cause you have onChange handler that changes that value in state. So it will redraw it.

src/App.tsx
```diff
@@ -20,17 +25,112 @@ class App extends React.Component<{}, State> {
       './img/9.png',
       './img/10.png',
     ],
+    step: 2,
```
sergii-nosachenko, line 28: Add type Readonly<State> to your state.

sergii-nosachenko (changes requested):
Something strange is going on sometimes:
1) when I change size from larger to smaller carousel doesn't redraw correctly [Screenshot](https://i.imgur.com/gIoP6Lu.png)
2) when I turn on infinite on the last or first item button stays disabled (but clickable - it's ok) [Screenshot](https://i.imgur.com/8M4HwSy.png)
3) the same when I turn off - button is active (but clickable - it's ok)
4) when I set step larger than images count arrow buttons become disabled but stay clickable [Screenshot](https://i.imgur.com/B6ICoo0.png)
5) fields accept strange values, you need to manage user input [Screenshot](https://i.imgur.com/yh5GoR8.png)
6) Your default values differs from those in task description.

denys-danyliuk (approved):
GJ!🚀

PR #212, opened 2022-09-18
Develop

AndreyJigijalkin:
[DEMO LINK](https://andreyjigijalkin.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,36 +1,132 @@
 import React from 'react';
 import './App.scss';
-import Carousel from './components/Carousel';
+import { Carousel } from './components/Carousel';
 
 interface State {
-  images: string[];
+  images?: string[];
+  itemWidth?: number;
+  step?: number;
+  frameSize?: number;
+  animationDuration?: number;
+  infinite?: boolean;
 }
+
+const smileImages = [
+  './img/1.png',
+  './img/2.png',
+  './img/3.png',
+  './img/4.png',
+  './img/5.png',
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,36 +1,150 @@
...
+    const { value } = event.target;
+
+    this.setState({ step: Number(value) });
+  };
+
+  handleAnimationDuration = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    this.setState({ animationDuration: Number(value) });
+  };
```
OlehMy, line 59: You can add attribute `name` to inputs and use only one `handleFunction`

OlehMy (changes requested):
Good work) But you should correct position when you are on the last element and increase `frameSize` ([here](https://prnt.sc/a525jNbO_56K))

OlehMy (approved):

src/App.tsx
```diff
@@ -1,36 +1,127 @@
 import React from 'react';
 import './App.scss';
-import Carousel from './components/Carousel';
+import { Carousel } from './components/Carousel';
+import { LabelInput } from './components/Label';
```
AlinaMatuschak, line 4: make sure that you pushed this file
Screenshot 2022-09-28 at 22 51 03

AlinaMatuschak (changes requested):
Awesome 👏 . Just check all files

PR #213, opened 2022-09-19
Solution without infinite loop

OlehMy:
[DEMO LINK](https://OlehMy.github.io/react_carousel/)

IMPORTANT!!!
The task is completed without infinite loop. You can just reset the position when the checkbox is true. But this is not the best solution, in my opinion. And a really correct infinite loop (like [this one](https://codepen.io/springerkc/pen/aMxxyO)) requires a bit more complex logic.

### DIFF
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
 
-// eslint-disable-next-line @typescript-eslint/ban-types
-type HandleFunction = Function;
+type HandleFunction = React.ChangeEventHandler<HTMLInputElement>;
 
 interface Props {
   label: string;
@@ -24,7 +23,6 @@ export const Input: React.FC<Props> = ({
 }) => {
   return (
     <div className="field">
-      {/* eslint-disable-next-line */}
       <label className="label">{label}</label>
       <div className="control">
         <input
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,9 +1,18 @@
...
+type Image = string;
 
 interface State {
-  images: string[];
+  images: Image[]
+  itemWidth: string;
+  frameSize: string;
+  step: string;
+  animationDuration: string;
+  // infinite: boolean;
```
dellachaise, line 14: Delete comments.

src/App.tsx
```diff
@@ -20,17 +29,118 @@ class App extends React.Component<{}, State> {
       './img/9.png',
       './img/10.png',
     ],
+    itemWidth: '130',
+    frameSize: '3',
+    step: '3',
+    animationDuration: '1000',
+    // infinite: false,
```
dellachaise, line 36: Here too.

src/App.tsx
```diff
@@ -20,17 +29,118 @@ class App extends React.Component<{}, State> {
...
+          : prevState.position,
+      }) as never;
+    });
+  };
+
+  // handleChecked = (event: React.ChangeEvent<HTMLInputElement>) => {
+  //   const { checked } = event.target;
+
+  //   this.setState({ infinite: checked });
+  // };
```
dellachaise, line 102: here the same.

dellachaise (approved):
Good job 👍 

Just delete comments all over the code.

src/App.tsx
OlehMy: Specially left. Maybe I'll implement infinite loop later))

PR #214, opened 2022-09-19
solution

matvii1:
https://matvii1.github.io/react_carousel/

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,26 @@
-// style go here
+@import url("https://fonts.googleapis.com/css?family=Montserrat+Subrayada:regular,700");
+
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
+}
+
+img {
+  display: block;
+}
+
+html {
+  font-family: "Montserrat", sans-serif;
+}
+
+.App h1 {
+  font-weight: 700;
+  width: max-content;
+  margin: 3rem auto 0;
+
+}
+
+ul {
+  list-style-type: none;
+}
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,88 @@
 import React from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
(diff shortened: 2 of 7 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,88 @@
...
+            });
+          }}
+        >
+          Prev
+        </button>
+        <button
+          type="button"
+          className="Next"
+          disabled={shiftX === disableRightPoint}
+          onClick={() => {
```
TarasHoliuk, line 74: this function looks bulky, please define it above and use here like `onClick={() => nextHandler()}`

src/components/Carousel.tsx
```diff
@@ -1,18 +1,88 @@
...
+                ? shiftX + imgWidth
+                : shiftX + imgWidth * shiftAmount),
+            });
+          }}
+        >
+          Prev
+        </button>
+        <button
+          type="button"
+          className="Next"
```
TarasHoliuk, line 72: not critical but would be nice to add styles to show if the button is disabled

src/components/Carousel.tsx
```diff
@@ -1,18 +1,88 @@
...
-export default Carousel;
+export class Carousel extends React.Component<Images, State> {
+  state = {
+    shiftX: 0,
+    imgWidth: 130,
+    shiftAmount: 3,
+    smallScrollRightPoint: -780,
+    smallScrollLeftPoint: -130,
+    disableRightPoint: -910,
+    disableLeftPoint: 0,
```
TarasHoliuk, line 26: if you change only `shiftX` (others are static) you can move all other props from the state (just declare as const)

src/App.scss
```diff
@@ -1 +1,26 @@
...
+
+img {
+  display: block;
+}
+
+html {
+  font-family: "Montserrat", sans-serif;
+}
+
+.App h1 {
```
TarasHoliuk, line 17: I guess you don't need a chained selector here

TarasHoliuk (changes requested):
Good job!
Left some comments

ArturErmolenko (approved):

PR #215, opened 2022-09-20
Develop

andredolgoleo:
[DEMO LINK](https://andredolgoleo.github.io/react_carousel/)


andredolgoleo:
The real solution is in 'fixed linter' commit. IDK how it happend

Bohdan-Maksymiuk (changes requested):
Good job 👍 
fix comments ⏬ 

Screenshot 2022-09-21 at 12 56 32

* images shouldn't be cropped after increasing sizes


https://user-images.githubusercontent.com/94289038/191475297-7f824250-8cec-4322-b53e-66ef843d5245.mov

* `Next` should became active after infinite mode turned on

andredolgoleo:
If it possible, tell me please, is it overcoded or it's a good solution?

Bohdan-Maksymiuk:
> If it possible, tell me please, is it overcoded or it's a good solution?

not bad)
More important that you became acquainted with refs in react 👍

PR #222, opened 2022-10-07
solution

ihorandrianov:
[DEMO LINK](https://ihorandrianov.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,106 @@
-import React from 'react';
-import './Carousel.scss';
+import React, { useEffect, useRef, useState } from 'react';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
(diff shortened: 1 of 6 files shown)


silchyn (changes requested):
image

src/components/Carousel.tsx
```diff
@@ -1,18 +1,105 @@
...
+        </button>
+        <div
+          className="overflow-hidden"
+          style={{
+            width: size * frameSize,
+          }}
+        >
+          <ul
+            className="Carousel__list w-full flex flex-nowrap"
+            style={{ translate: scrollTo, transitionDuration: `${animationDuration}ms` }}
```
naziamoff, line 79: a formatting of inline styles is not consistent

naziamoff (changes requested):
Nice job! We can improve it though.
- make the formatting of inline styles consistent
- disable/hide the button if there're no slides left

PR #224, opened 2022-10-14
solution

bohdan-tron:
[DEMO LINK](https://bohdan-tron.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,15 @@
 import React from 'react';
 import './App.scss';
-import Carousel from './components/Carousel';
+import { Carousel } from './components/Carousel';
 
-interface State {
-  images: string[];
-}
+type State = {
+  images: string[],
+  step: number,
+  frameSize: number,
+  itemWidth: number,
+  animationDuration: number,
+  infinite: boolean,
+};
 
 class App extends React.Component<{}, State> {
   state = {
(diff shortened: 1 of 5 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,148 @@
...
+      <div className="Carousel">
+        <ul className="Carousel__list" style={{ width: `${frameSize * itemWidth}px` }}>
+          {images.map((image, index) => {
+            return (
+              <>
+                <li
+                  className="Carousel__item"
+                  // eslint-disable-next-line react/no-array-index-key
+                  key={index}
+                  style={{
```
Bohdan-Maksymiuk, line 107: https://articles.wesionary.team/using-index-as-a-key-is-an-anti-pattern-in-react-8e5db3aea3a9

Bohdan-Maksymiuk (changes requested):
Good job 👍 
fix comment

src/App.tsx
```diff
@@ -1,36 +1,193 @@
...
     ],
+    itemWidth: 130,
+    frameSize: 3,
+    step: 3,
+    animationDuration: 1000,
+    infinite: false,
+  };
+
+  handlerOfInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const currenValue = +(event.currentTarget.value);
```
polosanya, line 68:
```suggestion
    const currentValue = +(event.currentTarget.value);
```

src/App.tsx
```diff
@@ -1,36 +1,193 @@
 import React from 'react';
 import './App.scss';
-import Carousel from './components/Carousel';
+import 'react-native-get-random-values';
+import { v4 as uuid } from 'uuid';
```
polosanya, line 4: You don't use `react-native-get-random-values` in your project and you didn't install `uuid` before using. 
If you want to generate random id, you can use default js methods [Math.random()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random) , [Date.now()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now), or even their combination

polosanya (changes requested):
Almost done! Please check my comments

src/App.tsx
bohdan-tron: зображення
here is the proof of working uuid from my demo-link, do you mean that I didnt upload uuid library to the commit?

src/App.tsx
```diff
@@ -1,36 +1,195 @@
...
+  step: number,
+  frameSize: number,
+  itemWidth: number,
+  animationDuration: number,
+  infinite: boolean,
+};
+
+function randomId() {
+  return String(Math.floor(Math.random() * 1000));
 }
```
polosanya, line 16: 🔥

polosanya (approved):
Well done!

PR #226, opened 2022-10-15
my solution

TheVovchik:
[DEMO LINK](https://TheVovchik.github.io/react_carousel/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
       infinite={false}
     />
     ```
-    
+
 ## REQUIREMENTS:
 
 1. The title of the page should contain "Carousel"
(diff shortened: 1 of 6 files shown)


silchyn (changes requested):
I think I found a bug

Reproduce steps:

- Stop the carousel
- Set the step to 10
- Try moving the carousel

TheVovchik:
> 

maybe because I need to awoid situations of 10 out of 10. so when you put this data carousel have only one key point (it's logic), so if you stop after change to 10, react recreate component and now it have only one keypoint = 0, so if I  tap next I haven't next point, but I was stopped not on the zero coordinate so prev button move me once to start point.
I remove posibility to choose step equal to image length to awoid this uncommon situation

PR #234, opened 2022-10-23
react carousel implemented

artem-musii:
[DEMO LINK](https://artem-musii.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,185 @@
-import React from 'react';
+import { Component } from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
-      <li><img src="./img/1.png" alt="3" /></li>
-      <li><img src="./img/1.png" alt="4" /></li>
-    </ul>
+interface Props {
+  images: string[];
+}
 
-    <button type="button">Prev</button>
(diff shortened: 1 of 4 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,186 @@
...
+                  transform: `translateX(-${position * itemWidth}px)`,
+                  transition: `transform ${animationDuration}s`,
+                }
+              }
+            >
+              {images.map((image, index) => (
+                <li key={image}>
+                  <img
+                    src={image}
+                    alt={`${index + 1}`}
```
naziamoff, line 93: the `alt` attribute should give us the alternative info. I understand it is not easy in this case, but let's try to come up with sth like that
```suggestion
                    alt={`emoji ${index + 1}`}
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,186 @@
...
+            step={0.1}
+            onChange={(event) => {
+              this.setState({
+                animationDuration: Number(event.currentTarget.value),
+              });
+            }}
+          />
+        </label>
+
+        <label className="Carousel__label">
```
naziamoff, line 171: I think it would be better to use `htmlFor` here in order not to wrap the input inside the label. But it is up to you to decide.

naziamoff (changes requested):
Good job! There are some comments though

- That is not really friendly interface for a user, I have to point at the item frame size selector each time after changing. Think of a way to correct this behavior

https://user-images.githubusercontent.com/103033758/197485666-d8f31ab5-e475-4fa3-adc7-402371e3c3a9.mov

naziamoff (approved):
🔥

PR #239, opened 2022-10-28
carousel was created

iPanda177:
[DEMO LINK](https://iPanda177.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,291 @@
 import React from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
-      <li><img src="./img/1.png" alt="3" /></li>
-      <li><img src="./img/1.png" alt="4" /></li>
-    </ul>
+type Props = {
+  images: string[];
+};
 
-    <button type="button">Prev</button>
-    <button type="button">Next</button>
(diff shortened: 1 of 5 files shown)


silchyn (changes requested):
Looks like bug

image

iPanda177:
tried a lot of things, but nothing works correctly (looks like its all about images resolution)
so i decide to go tricky way )

silchyn (changes requested):
Haha, not a solution

Denchepornyuk (changes requested):
https://user-images.githubusercontent.com/37452763/200279067-a83ce1a2-b869-45e1-9142-2adf3eb68a3d.mov

carousel shouldn't work like this

Anna-Tkachenko (changes requested):
when I choose infinite mode, carousel starts redirect me from the start to the end through all elements even if I chose step = 1, check it
image

iPanda177:
for fast look, i can't fix it quickly
so for now i've disabled this function and will dive deep later, when have time)

Infinite attribute was an extra task, so i hope you will approve this work without it 🙏

src/components/Carousel.tsx
```diff
@@ -1,18 +1,277 @@
...
+  disabledLeft: boolean;
+  disabledRight: boolean;
+};
+
+export class Carousel extends React.Component<Props, State> {
+  state = {
+    step: 3,
+    frameSize: 3,
+    itemWidth: 130,
+    animationDuration: 1000,
```
Anna-Tkachenko, line 24: will be great to move main values to some constants

Anna-Tkachenko (approved):
Good job! ⭐️

PR #240, opened 2022-11-01
done!

dmitriyn3679:
- [DEMO LINK](https://dmitriyn3679.github.io/react_carousel/)

### DIFF
--- a/src/components/ControlPanel.tsx
+++ b/src/components/ControlPanel.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './ControlPanel.scss';
+import { OptionInterface } from '../types/Option';
 
 type Props = {
   onChangeOptions: (option: string, value: number) => void;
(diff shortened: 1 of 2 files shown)


src/components/ControlPanel.tsx
```diff
@@ -0,0 +1,83 @@
...
+    {
+      title: 'Animation Duration',
+      name: 'animationDuration',
+      id: 'animationDuration',
+      defaultValue: defaultAnimationDuration,
+      min: '0',
+      max: '5000',
+      step: '500',
+    },
+  ];
```
Virass, line 60: Use class for build this array

Virass (changes requested):

src/components/Carousel.tsx
```diff
@@ -1,18 +1,84 @@
...
+) => {
+  const [imagePosition, setImagePosition] = useState(0);
+
+  const nextBtnLimit
+    = ((itemWidth * imageUrls.length) - (frameSize * itemWidth)) * -1;
+
+  const prevBtnLimit = 0;
+
+  const nextImg = () => {
+    setImagePosition(current => current - (step * itemWidth));
```
TarasHoliuk, line 29: you should also check remain space in the function to avoid the following:
image

TarasHoliuk (changes requested):
GJ! But one fix is needed

PR #242, opened 2022-11-07
first implementation

Blenemy:
[DEMO LINK](https://Blenemy.github.io/react_carousel/)

have an issue here. if i change the Itemwidth and then try to scroll, it shifts a bit, which was not intended...i suppose

### DIFF
--- a/src/components/Carousel.scss
+++ b/src/components/Carousel.scss
@@ -1,1 +1,186 @@
-// style go here
+* {
+  padding: 0;
+  margin: 0;
+  border: 0;
+}
+
+*,
+*:before,
+*:after {
+  -moz-box-sizing: border-box;
+  -webkit-box-sizing: border-box;
+  box-sizing: border-box;
+}
+
+:focus,
+:active {
+  outline: none;
+}
+
+a,
+a:visited {
+  text-decoration: none;
+}
+
+a:hover {
+  text-decoration: none;
+}
+
+a:focus,
+a:active {
+  outline: none;
+}
+
+nav,
+footer,
+header,
+aside {
+  display: block;
+}
+
+html,
+body {
+  height: 100%;
+  width: 100%;
+  line-height: 1;
+  font-size: 18px;
+  -ms-text-size-adjust: 100%;
+  -moz-text-size-adjust: 100%;
+  -webkit-text-size-adjust: 100%;
+}
+
+input,
+button,
+textarea {
+  font-family: inherit;
+}
+
+input::-ms-clear {
+  display: none;
+}
+
+button {
+  cursor: pointer;
+}
+
+button::-moz-focus-inner {
+  padding: 0;
+  border: 0;
+}
+
+ul li {
+  list-style: none;
+}
+
+img {
+  vertical-align: top;
+}
+
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+  font-size: inherit;
+  font-weight: 400;
+}
+
+body {
+  background: linear-gradient(#2ec771, #1abc9c);
+  padding: 20px;
+}
+
+.wrapper {
+  padding: 20px;
+
+  &__items {
+    display: flex;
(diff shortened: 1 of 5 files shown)


src/components/Carousel.scss
```diff
@@ -1 +1,186 @@
-// style go here
+* {
+  padding: 0;
+  margin: 0;
+  border: 0;
+}
+
+*,
+*:before,
+*:after {
```
qoa11a, line 9:
```suggestion
*::before,
*::after {
```

src/components/Carousel.scss
```diff
@@ -1 +1,186 @@
...
+  -ms-text-size-adjust: 100%;
+  -moz-text-size-adjust: 100%;
+  -webkit-text-size-adjust: 100%;
+}
+
+input,
+button,
+textarea {
+  font-family: inherit;
+}
```
qoa11a, line 56: There's a lot of styles that don't relate to carousel. Move them to separate file and leave here only styles needed for carousel.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,207 @@
...
+            disabled={transtaleX === 0}
+            onClick={() => {
+              if (transtaleX + translateShift > 0) {
+                this.setState({ transtaleX: 0 });
+
+                return;
+              }
+
+              this.setState({ transtaleX: transtaleX + translateShift });
+            }}
```
qoa11a, line 72: Move to a function.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,207 @@
...
+                - containerWidth <= -maxContainerWidth) {
+                this.setState({
+                  transtaleX: -maxContainerWidth + containerWidth,
+                });
+
+                return;
+              }
+
+              this.setState({ transtaleX: transtaleX + -translateShift });
+            }}
```
qoa11a, line 112: Move to a function.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,207 @@
...
+        <div className="wrapper__items">
+          <div className="wrapper__item wrapper__item-framesize">
+            <label htmlFor="">
+              {`Frame size is ${frameSize} `}
+              <input
+                onInput={(e) => {
+                  this.setState({
+                    frameSize: +e.currentTarget.value,
+                  });
+                }}
```
qoa11a, line 126: Move to a function.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,207 @@
...
+          </div>
+          <div className="wrapper__item wrapper__item-itemWidth">
+            <label htmlFor="">
+              {`Itemwidth is ${itemWidth}px `}
+              <input
+                onInput={(e) => {
+                  this.setState({
+                    itemWidth: +e.currentTarget.value,
+                  });
+                }}
```
qoa11a, line 158: Move to a function.

qoa11a (changes requested):

src/components/Carousel.tsx
Blenemy: didn't move to a function this event cause i wanted to use variables.

src/components/Carousel.tsx
Blenemy: didn't move to a function this event cause i wanted to use variables.

qoa11a (changes requested):

Blenemy:
wanted to know btw how to fix the problem which i described at the top. thank you for your comments. looking forward to hearing from you

Anastasiia-Svintsova (changes requested):
- You have the issue with shift because you have the wrong container size. Every item has a width of 200 px, but the container is 390 for example. Check this
image
- Infinite scroll means that I can endlessly scroll through elements, and not automatic animation
Animation

qoa11a (approved):

PR #243, opened 2022-11-09
React_carousek

DenysKleimenov:
[DEMO LINK](https://DenysKleimenov.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,93 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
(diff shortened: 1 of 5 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,93 @@
...
+        <div
+          className="container2"
+          style={{
+            transform: `translateX(${movePx}px)`,
+            transition: `transform ${animationDuration}ms`,
+          }}
+        >
+          <ul className="Carousel__list">
+            {images.map((image, index) => (
+              <li key={image}>
```
anastasiia-nahaiska, line 59: Maybe would be better to create an array with images object with path and id properties and don't use a image as a key

anastasiia-nahaiska (approved):
looks cool!

src/components/Carousel.tsx
DenysKleimenov: Thanks for review 🤗

silchyn (changes requested):
image

PR #250, opened 2022-11-22
slider done

Alina-Yermolenko:
[DEMO LINK](https://Alina-Yermolenko.github.io/react_carousel/)


Anastasiia-Svintsova (changes requested):
There is an issue with a `Prev` button. I think you should disable it if there are no previous items.
Animation

src/App.scss
```diff
@@ -1 +1,9 @@
 // style go here
+.input {
+    display: flex;
+
+    &-block {
+        display: flex;
+        flex-direction: column;
+    }
+}
```
loralevitska, line 9: add empty line after end in each file

src/App.tsx
```diff
@@ -20,17 +24,107 @@ class App extends React.Component<{}, State> {
...
-        <h1>Carousel with {images.length} images</h1>
+        <h1 data-cy={'title'}>Carousel with {images.length} images</h1>
+        <div className="input">
+          <label className="input-block" htmlFor="itemId">
+            Item width
+            <input
+              value={this.state.itemWidth || 130}
+              onChange={(event) => {
+                return this.updateInputValue(event.target.value, 'itemWidth');
+              }}
```
loralevitska, line 61: Move to separate function  (transfer the function to a variable, and here only write the name of the function). Please fix all cases if present

src/components/Carousel.scss
```diff
@@ -1 +1,10 @@
 // style go here
+.Carousel {
```
loralevitska, line 2: class names must be in lowercase

src/components/Carousel.tsx
```diff
@@ -1,18 +1,122 @@
...
+
+class Carousel extends React.Component<Props, State> {
+  state = {
+    itemOffset: 0,
+  };
+
+  clickNext = () => {
+    const {
+      itemWidth, images, infinite, frameSize,
+    } = this.props;
```
loralevitska, line 24: Please fix all cases if present

```suggestion
    const {
      itemWidth,
      images,
      infinite,
      frameSize,
    } = this.props;
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,122 @@
...
+    const {
+      itemWidth, images, frameSize, animationDuration,
+    } = this.props;
+
+    return (
+      <div
+        className="Carousel"
+        style={{ width: `${itemWidth * frameSize}px` }}
+      >
+
```
loralevitska, line 82:
```suggestion
```

loralevitska (changes requested):
1. The problem remained. You need to limit the container (with a large number of pictures) as you limit with a value of 3 pictures (and everything works correctly)

https://user-images.githubusercontent.com/70009205/205430028-8373b74e-0500-4eed-9aa5-5b7f8e232b45.mov

PR #252, opened 2022-11-29
Carousel task solution

juliiaap:
- [DEMO LINK](https://juliiaap.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -130,14 +130,12 @@ class Carousel extends React.Component<Props> {
         >
           {images.map((image) => (
             <li className="Carousel__item" key={image}>
-              <div className="image-container" style={{ width: `${itemWidth}px` }}>
-                <img
-                  src={image}
-                  alt={image}
-                  className="Carousel__image"
-                  width={itemWidth}
-                />
-              </div>
+              <img
+                src={image}
+                alt={image}
+                className="Carousel__image"
+                width={itemWidth}
+              />
             </li>
           ))}
         </ul>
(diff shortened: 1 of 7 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,167 @@
...
+          behavior: 'smooth',
+        });
+      }, 2000);
+    }
+
+    if (Math.ceil(scrollLeft) === (scrollWidth
+      - clientWidth)) {
+      nextButton?.classList.add('disabled');
+    } else if (scrollLeft === 0) {
+      prevButton?.classList.add('disabled');
```
polosanya, line 105: Else if looks complicated. Try using switch...case instead or\and ternary operator

src/components/Carousel.tsx
```diff
@@ -1,18 +1,167 @@
...
+      itemWidth,
+    } = this.props;
+    const containerWidth = frameSize * (itemWidth + gap) - gap;
+
+    return (
+      <div className="Carousel">
+        <ul
+          className="Carousel__list"
+          style={{ width: `${containerWidth}px` }}
+          onScroll={(e) => {
```
polosanya, line 125: Move it to the separate function

polosanya (changes requested):
Good job! Please check the comments to improve your code a bit

src/components/Carousel.tsx
juliiaap: rewrote with switch...case

src/components/Carousel.tsx
juliiaap: done

polosanya (approved):
Well done!

PR #261, opened 2022-12-14
done

Nazar-Brunarskyi:
[DEMO LINK](https://Nazar-Brunarskyi.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,8 @@
 import React from 'react';
 import './App.scss';
-import Carousel from './components/Carousel';
-
-interface State {
-  images: string[];
-}
+import './Page/Page.scss';
+import { Carousel } from './components/Carousel';
+import { Carousel as State } from './types/Carousel';
 
 class App extends React.Component<{}, State> {
   state = {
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -20,17 +18,217 @@ class App extends React.Component<{}, State> {
...
+        break;
+
+      default:
+        newValue = curentValue;
+        break;
+    }
+
+    this.setState({
+      step: newValue,
+    });
```
romanblyashuk94, line 48: it can be shorter)
fix in all set functions

```suggestion
    let curentValue = Number(event.currentTarget.value);

    switch (true) {
      case curentValue > 10:
        this.setState({ step: 10 });
        break;
      case curentValue < 0:
        this.setState({ step: 0 });
        break;

      default:
        this.setState({ step: curentValue });
    }
```

src/App.tsx
```diff
@@ -20,17 +18,217 @@ class App extends React.Component<{}, State> {
...
+          <div className="Page__Input-block">
+            <label className="Page__input-lable" htmlFor="itemId">
+              {'Item Width '}
+            </label>
+
+            <input
+              className="Page__input"
+              id="itemId"
+              type="number"
+              defaultValue={itemWidth}
```
romanblyashuk94, line 189: use "value" attribute here and other inputs
UI must be dependent on state

src/components/Carousel.scss
```diff
@@ -1 +1,48 @@
...
+    width: #{$i}px;
+    height: #{$i}px;
+  }
+}
+
+@for $i from 0 through 10000 {
+  .Animation-duration-#{$i} {
+    transition: transform #{$i}ms ease;
+  }
+}
```
romanblyashuk94, line 48: you generate here 61300 selectors))))

that works, but I think using inline styles  in markup based on props is better way ;)

romanblyashuk94 (changes requested):

romanblyashuk94 (approved):

src/App.tsx
```diff
@@ -20,17 +18,193 @@ class App extends React.Component<{}, State> {
...
+    itemWidth: 130,
+    animationDuration: 1000,
+    infinite: false,
+  };
+
+  setStep = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const curentValue = Number(event.currentTarget.value);
+
+    switch (true) {
+      case curentValue > 10:
```
solaryasha, line 32: What is this magic number ?

src/App.tsx
```diff
@@ -20,17 +18,193 @@ class App extends React.Component<{}, State> {
...
+        this.setState({ step: 0 });
+        break;
+
+      default:
+        this.setState({ step: curentValue });
+        break;
+    }
+  };
+
+  setItemWidth = (event: React.ChangeEvent<HTMLInputElement>) => {
```
solaryasha, line 45: Do you really need four different methods ?
Maybe, let's write one that uses dynamic object key.

src/App.tsx
```diff
@@ -20,17 +18,193 @@ class App extends React.Component<{}, State> {
...
+    const curentValue = Number(event.currentTarget.value);
+
+    switch (true) {
+      case curentValue > 1300:
```
solaryasha, line 49: What is this magic number ?

src/App.tsx
```diff
@@ -20,17 +18,193 @@ class App extends React.Component<{}, State> {
...
+        this.setState({ frameSize: curentValue });
+        break;
+    }
+  };
+
+  setAnimationDuration = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const curentValue = Number(event.currentTarget.value);
+
+    switch (true) {
+      case curentValue > 10000:
```
solaryasha, line 83: What is this magic number ?

src/App.tsx
```diff
@@ -20,17 +18,193 @@ class App extends React.Component<{}, State> {
...
+            <label className="Page__input-lable" htmlFor="stepId">
+              {'Step '}
+            </label>
+
+            <input
+              className="Page__input"
+              id="stepId"
+              type="number"
+              value={step}
+              onChange={(event) => this.setStep(event)}
```
solaryasha, line 134:
```suggestion
              onChange={this.setStep}
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,158 @@
...
+                  className="Carousel__list-item"
+                  style={{
+                    width: itemWidth,
+                    height: itemWidth,
+                  }}
+                >
+                  <img
+                    className="Carousel__item-img"
+                    src={image}
+                    alt={`${index + 1}`}
```
solaryasha, line 128: It's very informative :)
We should use better descriptions

solaryasha (changes requested):

solaryasha (approved):
👍

PR #263, opened 2022-12-14
Initial

UmizDemud:


src/App.tsx
```diff
@@ -1,39 +1,64 @@
...
-  };
+  return (
+    <div className="App">
+      <div>
+        <h1
+          className="title"
+          data-cy="title"
+        >
+          {/* eslint-disable-next-line */}
+          Carousel with {images.length} images
```
meforoma, line 33: better avoid disabling linter
i think below should make linter happy
```suggestion
          {`Carousel with ${images.length} images`}
```

src/App.tsx
```diff
@@ -1,39 +1,64 @@
...
+        itemWidth={itemWidth}
+        animationDuration={animationDuration}
+        infinite={infinite}
+        sS={{
+          step: setStep,
+          frameSize: setFrameSize,
+          itemWidth: setItemWidth,
+          animationDuration: setAnimationDuration,
+          infinite: setInfinite,
+        }}
```
meforoma, line 58: it's kind of reasonable idea from structuring perspective, but it often causes issues accessing nested props - i'd rather pass them with distinct naming:
```suggestion
        setStep: setStep,
        setFrameSize: setFrameSize,
        setItemWidth: setItemWidth,
        setAnimationDuration: setAnimationDuration,
        setInfinite: setInfinite,
```

or at least provide a meaningful nested prop name:
```suggestion
        stateSetters={{
          step: setStep,
          frameSize: setFrameSize,
          itemWidth: setItemWidth,
          animationDuration: setAnimationDuration,
          infinite: setInfinite,
        }}
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,276 @@
...
+ * A function to determine where the incoming items should be directed from.
+ * Let's see if it will be implemented.
+  setPositions(towards: boolean) {
+    const {
+      frameSize,
+      step,
+      images,
+    } = this.props;
+
+    const {
```
meforoma, line 242: don't leave the commented code in committed files

src/components/Options/Input/Input.tsx
```diff
@@ -27,28 +27,28 @@ export const Input = ({ props }: Props) => {
...
 
     if (name === 'Step') {
-      if (!val || (val >= 0 && val <= 5)) {
+      if (val >= 0 && val <= 5) {
         setError(false);
         handleChange(e);
       } else {
         setError(true);
       }
     } else if (name === 'Frame size') {
```
meforoma, line 36: not sure, but feels like all `else if` could be replaced with a `switch..case` construct?

src/components/Options/Input/Input.tsx
```diff
@@ -27,28 +27,28 @@ export const Input = ({ props }: Props) => {
...
         setError(false);
         handleChange(e);
       } else {
         setError(true);
       }
     } else if (name === 'Animation duration') {
-      if (!val || val >= 0) {
+      if (val >= 0) {
         setError(false);
         handleChange(e);
```
meforoma, line 53: like
```suggestion
    case 'Item width':
    case 'Animation duration':
      if (val >= 0) {
        setError(false);
        handleChange(e);
      } else {
        setError(true);
      }
    case...
    default...
```

src/components/Options/Input/Input.tsx
```diff
@@ -0,0 +1,102 @@
...
+          checked={value as boolean}
+          type="checkbox"
+          name="inf"
+          id={`${id}Id`}
+        />
+      </label>
+    );
+  }
+
+  return <></>;
```
meforoma, line 101: essentially `null` will render nothing, a general practice
```suggestion
  return null;
```

src/components/Options/Input/Input.tsx
```diff
@@ -0,0 +1,102 @@
...
+        :
+        <input
+          className="Options__input"
+          onChange={(e) => dimensionChecks(e)}
+          placeholder={value.toString()}
+          type={type}
+          name={name}
+          id={`${id}Id`}
+        />
+      </label>
```
meforoma, line 78: consider if likes of this approach would do the job
to achieve avoiding conditional returns

```
const errorClass = error ? ' Options__error' : '';
const className = type === 'number'
  ? `Options__label${errorClass}`
  : 'Options__label';

const label = (
	<label
     htmlFor={`${id}Id`}
     className={className}
  >
)

const input = type === 'number'
  ? (
      <input
          className="Options__input"
          onChange={(e) => dimensionChecks(e)}
          placeholder={value.toString()}
          type={type}
          name={name}
          id={`${id}Id`}
      />
  ) : (
    <input
          className="Options__input"
          onChange={(e) => dimensionChecks(e)}
          checked={value as boolean}
          type="checkbox"
          name="inf"
          id={`${id}Id`}
        />
  )

  return ['boolean', 'number'].includes(type) && (
		{label}
		{name}
		:
		{input}
  )
```

meforoma (changes requested):
@UmizDemud let's evaluate whether my suggestions would do any better.
I hope they will)

src/components/Options/Input/Input.tsx
UmizDemud: For sure, would look much more understandable

src/components/Options/Input/Input.tsx
UmizDemud: 👍

src/components/Options/Options.tsx
```diff
@@ -38,40 +40,45 @@ export const Options = (props: Props) => {
       name: 'Step',
       id: 'step',
       value: step,
-      handleChange: (e: React.ChangeEvent<HTMLInputElement>) => sS
-        .step(Number.parseInt(e.target.value, 10)),
+      handleChange: (e: React.ChangeEvent<HTMLInputElement>) => setStep(
+        Number.parseInt(e.target.value, 10),
+      ),
```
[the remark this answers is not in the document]
UmizDemud: I kinda made some way
```js
  const changeHandler = (
    stateSetter: React.Dispatch<React.SetStateAction<number | boolean>>,
    e: React.ChangeEvent<HTMLInputElement>,
    isOutputBoolean = false,
  ) => stateSetter(isOutputBoolean
    ? Boolean(e.target.checked)
    : Number.parseInt(e.target.value, 10));

  const inputs:{
    type: 'number' | 'boolean',
    name: string,
    id: string,
    value: number | boolean,
    handleChange: typeof changeHandler,
....
```

like this. But then I have to pass the function and multiple state setters. Or it will be more lines than what I started with. Am I right? 

Also I'm using useState hook and 5-6 variables

src/components/Options/Input/Input.tsx
UmizDemud: 👍

src/components/Options/Input/Input.tsx
UmizDemud: Implemented this, it looks better imo. Just had to put a 
```|| null```
at the end because apparently these components can't return false

meforoma (approved):

PR #270, opened 2022-12-25
react_carousel

ysrckr:
[DEMO LINK](https://ysrckr.github.io/react_carousel/)


src/App.tsx
```diff
@@ -20,17 +29,138 @@ class App extends React.Component<{}, State> {
...
+  };
+
+  stepChangeHandler = (event: ChangeEvent<HTMLInputElement>) => {
+    if (event.target.value === ''
+      || Number(event.target.value) < 1
+      || Number.isNaN(Number(event.target.value))) {
+      this.setState({
+        step: 1,
+      });
+    }
```
meforoma, line 46: thinking, would this entire evaluation be even needed if you set the input `type="number"`?
i believe not - so let's go for typing the inputs?

src/App.tsx
```diff
@@ -20,17 +29,138 @@ class App extends React.Component<{}, State> {
...
+  frameSizeChangeHandler = (event: ChangeEvent<HTMLInputElement>) => {
+    if (
+      event.target.value === ''
+      || Number(event.target.value) < 1
+      || Number.isNaN(Number(event.target.value))
+    ) {
+      this.setState({
+        frameSize: 1,
+      });
+    }
```
meforoma, line 62: same here - typed input (type="number") would make this not needed

src/App.tsx
```diff
@@ -20,17 +29,138 @@ class App extends React.Component<{}, State> {
...
+
+    this.setState({
+      frameSize: Number(event.target.value),
+    });
```
meforoma, line 66: you could also think of this approach - unify all the numeric inputs change handler - so you only have just one input and that's it
```suggestion
    const { name, value } = event.target;

    this.setState({
      [name]: Number(value),
    });
```

to make this work you'll need to provide each Input with a name attribute that corresponds with the state names.
```
<input type="number" name="frameSize" ...
// and so on
```

meforoma (changes requested):
decent 👍🏼 
please see the comments i wrote - let's improve the implementation iteratively

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,95 @@
...
+  };
+
+  const listStyle = {
+    width: `${(itemWidth * frameSize)}px`,
+  };
+
+  const handleNext = () => {
+    const length = Math.ceil(images.length / (frameSize * step));
+
+    if (infinite
```
naziamoff, line 36: better to separate conditions like that to a variable

naziamoff (approved):
well done 👍 🔥

meforoma (approved):

PR #273, opened 2022-12-27
dom_carousel

kolya-movchan:
[DEMO LINK](https://kolya-movchan.github.io/react_carousel/)


src/App.tsx
```diff
@@ -22,15 +34,121 @@ class App extends React.Component<{}, State> {
...
+        className="input"
+        type="number"
+        value={animationDuration}
+        name="animationDuration"
+        onChange={(event) => this.handleChange(event, 'animationDuration')}
+        min="0"
+      />
+    );
+
+    const infiniteInput = (
```
TarasHoliuk, line 112: Not sure it's clearer to define inputs as variables. You don't reuse them, so IMO it's better to write them on the place

The same with `emojis` in `Carousel.tsx`

TarasHoliuk (changes requested):
GJ :+1: Nice background :smile: 

Try to get rid of scroll on the page

TarasHoliuk (approved):
Amazing :fire:

PR #278, opened 2023-01-01
Develop

nurdanokcu:
[DEMO LINK](https://nurdanokcu.github.io/react_carousel/)


src/App.tsx
```diff
@@ -20,20 +26,64 @@ class App extends React.Component<{}, State> {
...
+    frameSize: 3,
+    itemWidth: 130,
+    animationDuration: 1000,
+    infinite: false,
+  };
+
+  handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    this.setState(state => ({
+      ...state,
+      [event.target.name]: +(event.target.value),
```
meforoma, line 39: nice, but let's convert data types explicitly
```suggestion
      [event.target.name]: Number(event.target.value),
```

nurdanokcu: Thank you Roman.  I added it now.

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,92 @@
...
+        className="Form__input"
+        type="number"
+        name="itemWidth"
+        value={itemWidth}
+        min="100"
+        max="200"
+        step="5"
+        onChange={handleChange}
+      />
+    </label>
```
meforoma, line 38: i think you could create an `Input.tsx` component - although it might be bit cumbersome
yet if you have some spare time - would be a good practice with components.

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,120 @@
...
+        </ul>
+
+        <div className="Carousel__buttons">
+          <button
+            type="button"
+            className="Carousel__button"
+            onClick={this.handlePrevButton}
+            disabled={transform === 0 && !infinite}
+          >
+            &#8249;
```
meforoma, line 102: i'd prefer a string to not look for the charcode 🙂 
```suggestion
            {'<'};
```

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,120 @@
...
+          <button
+            data-cy="next"
+            type="button"
+            className="Carousel__button"
+            onClick={this.handleNextButton}
+            // calculated size is sizeOfHiddenRegion
+            disabled={transform === -(images.length - frameSize) * itemWidth
+              && !infinite}
+          >
+            &#8250;
```
meforoma, line 114: i'd prefer a string to not look for the charcode 🙂 
```suggestion
            {'>'};
```

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,92 @@
...
+  handleChange,
+  frameSize,
+  step,
+  animationDuration,
+  images,
+  infinite,
+  handleInfinite,
+}) => (
+  <form className="Form">
+    <label>
```
meforoma, line 26: let's satisfy test by adding `<label for...` and `<input id...`

meforoma (changes requested):
indeed the implementation is functional.
Would be nice to polish it up regarding tests and minor refactorings.
And i'd prefer that you have `Input.tsx` component which will render the relevant input based on passed props.

@nurdanokcu regarding the tests - you can skim through the file 
`cypress/integration/page.spec.js` - and you'll find what it tries to do

among others i see tests are trying to find labels by `for` attribute
failing tests
so you can figure out what's going wrong about tests.

meforoma (approved):
good job 🔥

PR #280, opened 2023-01-02
solution

lefteris-L:
[DEMO LINK](https://lefteris-l.github.io/react_carousel/)


src/components/Inputs/Inputs.tsx
```diff
@@ -0,0 +1,153 @@
...
+        <input
+          type="number"
+          id="stepId"
+          value={step}
+          min={1}
+          max={maxByLength}
+          className="inputs__field"
+          onChange={handleChange}
+        />
+      </label>
```
meforoma, line 87: would be really nice to make an `Input.tsx` component that would render the desired input based on passed props

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+      <button
+        className={cn('button next', {
+          '--disabled': isDisabledNext,
+        })}
+        type="button"
+        data-cy="next"
+        onClick={handleNext}
+      >
+        &nbsp;&gt;&nbsp;
```
meforoma, line 164: not cool ))
let's better use a string literal to read this without conversion
and the same applies for `prev`
```suggestion
        {` > `};
```

src/types.ts
```diff
@@ -0,0 +1,20 @@
...
+  itemWidth: number
+  animationDuration: number
+  infinite: boolean
+  distance: number
+}
+
+export type Hook<T> = React.Dispatch<React.SetStateAction<T>>;
+
+export interface InputHooks {
+  setStep: Hook<number>,
```
meforoma, line 14: nice type simplification 👍🏼

meforoma (changes requested):
please see the comments i've left - let's apply those.

meforoma (approved):
awesome 👏🏼

PR #297, opened 2023-01-31
react_carousel

rafmamedov:
[DEMO LINK](https://rafmamedov.github.io/react_carousel/)


src/App.tsx
```diff
@@ -20,20 +26,63 @@ class App extends React.Component<{}, State> {
...
+      animationDuration,
+      infinite,
+    } = this.state;
 
     return (
       <div className="App">
         {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1 data-cy="title">
+          {`Carousel with ${images.length} images`}
```
Volodymyr-Melnyk-Fron10der, line 63:
```suggestion
          Carousel with {images.length} images
```

src/App.tsx
```diff
@@ -20,20 +26,63 @@ class App extends React.Component<{}, State> {
...
       './img/10.png',
     ],
+    step: 3,
+    frameSize: 3,
+    itemWidth: 130,
+    animationDuration: 1000,
+    infinite: false,
+  };
+
+  handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
Volodymyr-Melnyk-Fron10der, line 36: Write methods with correct syntax
```suggestion
  handleChange(event: React.ChangeEvent<HTMLInputElement>) {
```

src/App.tsx
```diff
@@ -20,20 +26,63 @@ class App extends React.Component<{}, State> {
...
+    this.setState(state => ({
+      ...state,
+      [event.target.name]: +(event.target.value),
+    }));
+  };
+
+  handleInfinite = () => {
```
Volodymyr-Melnyk-Fron10der, line 43:
```suggestion
  handleInfinite() {
```

Volodymyr-Melnyk-Fron10der (changes requested):
Great, just some fixes:

src/App.tsx
rafmamedov: Implemented this, thank you!

src/App.tsx
rafmamedov: I can't do that, because if the function is not an arrow function, 'this' inside it is defined as undefined, so the method doesn't work, and some tests crash.

SanyaBratashchuk (approved):

PR #308, opened 2023-02-16
solution

TetianaKoval:
[DEMO LINK](https://TetianaKoval.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.scss
+++ b/src/components/Carousel.scss
@@ -1,1 +1,21 @@
 // style go here
+.Carousel {
+  overflow: hidden;
+
+  &__list {
+    margin: 0;
+    padding: 0;
+    width: 1300px;
+    display: flex;
+    flex-direction: row;
+
+    & li {
+      list-style: none;
+
+      & img {
+        width: inherit;
+        height: inherit;
+      }
+    }
+  }
+}
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,123 @@
 import React from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
-      <li><img src="./img/1.png" alt="3" /></li>
-      <li><img src="./img/1.png" alt="4" /></li>
-    </ul>
+type State = {
+  position: number,
+};
 
-    <button type="button">Prev</button>
-    <button type="button">Next</button>
-  </div>
-);
+type Props = {
+  images: string[],
+  step: number,
+  frameSize: number,
+  itemWidth: number,
+  animationDuration: number,
+  infinite: boolean,
+};
+
+class Carousel extends React.Component<Props, State> {
+  state = {
+    position: 0,
+  };
+
+  moveSlider = (buttonName: string) => {
+    const {
+      images,
+      step,
+      frameSize,
+      itemWidth,
+      infinite,
+    } = this.props;
+    const imagesLength = itemWidth * images.length;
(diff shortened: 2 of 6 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,125 @@
...
+          className="Carousel__list"
+          style={{
+            transform: `translate(-${this.state.position}px)`,
+            transition: `transform ${animationDuration}ms ease`,
+          }}
+        >
+          {images.map((image, index) => (
+            <li
+              key={image}
+            >
```
anstsot, line 92: it would be better to write in one line
```suggestion
            <li key={image}>
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,125 @@
...
+  animationDuration: number,
+  infinite: boolean,
+};
+
+class Carousel extends React.Component<Props, State> {
+  state = {
+    position: 0,
+  };
+
+  moveSlider = (buttonName: string) => {
```
anstsot, line 22: you can add specific typing for `buttonName`:
```
enum ControlButton {
  Prev,
  Next,
}
```
and use it in `moveSlider` function
```suggestion
  moveSlider = (buttonName: ControlButton) => {
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,125 @@
...
+                height={itemWidth}
+              />
+            </li>
+          ))}
+        </ul>
+
+        <button
+          type="button"
+          onClick={(event) => {
+            this.moveSlider(event.currentTarget.innerText);
```
anstsot, line 106: then here it would be more convenient to use same typing `ControlButton`
```suggestion
            this.moveSlider(ControlButton.Prev);
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,125 @@
...
+  itemWidth: number,
+  animationDuration: number,
+  infinite: boolean,
+};
+
+class Carousel extends React.Component<Props, State> {
+  state = {
+    position: 0,
+  };
+
```
anstsot, line 21: maybe it would be worth to consider using the `getDerivedStateFromProps` and update position by changing some props?

anstsot (changes requested):
Looks good in general but some improvements are needed. Could you check the tests as well?

src/components/Carousel.tsx
```diff
@@ -1,18 +1,161 @@
...
+    currentStartImg: 0,
+  };
+
+  static getDerivedStateFromProps(props:Props, state:State) {
+    return {
+      position: state.currentStartImg * props.itemWidth,
+    };
+  }
+
+  moveSlider = (buttonName: ControlButton) => {
```
Volodymyr-Melnyk-Fron10der, line 35: Wouldn't be better to split logic for ControlButton.Prev & ControlButton.Next into separated functions? It looks a little hard to read and doesn't meet the refactoring patterns

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! 

1. right button should scroll as well when infinite checked 
image

src/components/Carousel.scss
```diff
@@ -1 +1,73 @@
 // style go here
```
Volodymyr-Melnyk-Fron10der, line 1: don't commit any comments

src/components/Carousel.scss
```diff
@@ -1 +1,73 @@
...
+        height: 0;
+        border-top: 8px solid transparent;
+        border-bottom: 8px solid transparent;
+        display: block;
+        border-right: 10px solid rgba(255, 255, 255, 0.37);
+        transition: all 300ms ease;
+      }
+
+      &:hover {
+        &::after {
```
Volodymyr-Melnyk-Fron10der, line 67:
```suggestion
      &:hover::after {
```

Volodymyr-Melnyk-Fron10der (approved):
Ok👍

PR #315, opened 2023-02-23
react_carousel

dima-korovai:
[DEMO LINK](https://dima-korovai.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -115,10 +115,10 @@ class Carousel extends Component<Props, State> {
         this.setState({ frameSize: +value });
         break;
       case 'imageSize':
-        this.setState({ frameSize: +value });
+        this.setState({ itemWidth: +value });
         break;
       case 'moutionSpeed':
-        this.setState({ frameSize: +value });
+        this.setState({ animationDuration: +value });
         break;
       case 'infinite':
         this.setState({ infinite: !infinite });


polosanya (changes requested):
Moving steps setting doesn't work correctly. Check yourself with different steps/frame sizes/infinity etc.
Step
steps

dima-korovai:
Thanks, seems everything to be fixed

polosanya (changes requested):
1. Why the last image is always cropped? Please fix it
image
2. First click made 4 steps, and the second is only 3
steps

dima-korovai:
About first comment, I thought if one pictures doesn"t reflect fully, it easier to understand that something lefts.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,245 @@
...
+      imagesAmount,
+      itemWidth,
+      movedDistance,
+      gap,
+      step,
+      infinite,
+      frameSize,
+    } = this.state;
+
+    if (infinite === true) {
```
agniya-i, line 47:
```suggestion
    if (infinite) {
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,245 @@
...
+    } else {
+      this.setState({
+        movedDistance: movedDistance - movedDistance,
+      });
+    }
+
+    const totalWidth = imagesAmount * (itemWidth + gap);
+    const leftWidth = (totalWidth - (frameSize * (itemWidth + gap)));
+
+    if ((infinite === true) && movedDistance === 0) {
```
agniya-i, line 96:
```suggestion
    if (infinite && movedDistance === 0) {
```

src/components/Carousel.scss
```diff
@@ -1 +1,81 @@
...
+      &::after {
+        content: "\25ba";
+        padding-left: 0.5em;
+      }
+    }
+  }
+
+  &__form {
+    position: absolute;
+    background-color: rgba(255, 174, 0, 0.59);
```
agniya-i, line 49: don't forget that you can create variables for repeated values ( colors in this case)

src/App.tsx
```diff
@@ -28,9 +39,21 @@ class App extends React.Component<{}, State> {
...
+        </h1>
 
-        <Carousel />
+        <Carousel
+          images={images}
+          step={this.state.step}
+          frameSize={this.state.frame}
+          itemWidth={this.state.itemWidth}
+          animationDuration={this.state.moveSpeed}
+          infinite={this.state.infinite}
```
agniya-i, line 55: use destructuring for `this.state`, to not repeat it each time

src/components/Carousel.tsx
```diff
@@ -1,18 +1,245 @@
...
+            {'Infinite: '}
+            <input
+              className="Carousel__form__input Carousel__form__input--checkbox"
+              type="checkbox"
+              onClick={() => {
+                if (infinite === false) {
+                  this.setState({ infinite: true });
+                } else {
+                  this.setState({ infinite: false });
+                }
```
agniya-i, line 234:
```suggestion
              
                  this.setState((prevState) => { ...prevState, infinite: !prevState.infinite });
         
```
you can do something like this, or try to set  `event.currentTarget.checked` value

src/components/Carousel.tsx
```diff
@@ -1,18 +1,245 @@
...
+            className="Carousel__button Carousel__button--prev"
+            type="button"
+            onClick={this.handleImagesPrev}
+          >
+            Prev
+          </button>
+
+          <button
+            data-cy="next"
+            disabled={movedDistance >= ((imagesAmount - frameSize)
```
agniya-i, line 157: it would be better to create separate variable for this value

agniya-i (changes requested):
Good job!
Let's improve a little bit

src/components/Carousel.tsx
dima-korovai: Ok. Thank you very much!

src/components/Carousel.tsx
```diff
@@ -1,18 +1,245 @@
...
+  gap: number,
+  frameSize: number,
+  step: number,
+  animationDuration: number,
+  infinite: boolean,
+};
+
+class Carousel extends React.Component<Props, State> {
+  state = {
+    imagesAmount: this.props.images.length,
```
Anastasiia-Svintsova, line 26: Would be good to destructure the props

src/components/Carousel.tsx
```diff
@@ -1,18 +1,245 @@
...
+    if (infinite) {
+      this.setState({ movedDistance: 0 });
+    }
+
+    const totalWidth = imagesAmount * (itemWidth + gap);
+
+    if ((movedDistance + frameSize * (itemWidth + gap))
+      < (imagesAmount * (itemWidth + gap))) {
+      const leftWidth = (totalWidth - (movedDistance
+        + (frameSize * (itemWidth + gap))));
```
Anastasiia-Svintsova, line 56: It's really hard to read, you could move some repeatable values to variables. `frameSize * (itemWidth + gap)` for example. And you already have a variable for `imagesAmount * (itemWidth + gap)`, so, you could use it

src/components/Carousel.tsx
```diff
@@ -1,18 +1,245 @@
...
+          {images.map((image, index) => {
+            return (
+              <li
+                className="Carousel__item"
+                key={image}
+                style={{ marginRight: `${gap}px` }}
+              >
+                <img
+                  src={image}
+                  alt={`${index + 1}`}
```
Anastasiia-Svintsova, line 140: Strange alt name. It should be more meaningful

src/components/Carousel.tsx
```diff
@@ -1,18 +1,245 @@
...
+                  width={`${itemWidth}`}
+                />
+              </li>
+            );
+          })}
+        </ul>
+
+        <div className="Carousel__buttons">
+          <button
+            disabled={movedDistance === 0 && infinite === false}
```
Anastasiia-Svintsova, line 150:
```suggestion
            disabled={!movedDistance && !infinite}
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,245 @@
...
+            type="button"
+            onClick={this.handleImagesPrev}
+          >
+            Prev
+          </button>
+
+          <button
+            data-cy="next"
+            disabled={movedDistance >= lastFrameSizeWidth
+              && infinite === false}
```
Anastasiia-Svintsova, line 161: Use logic operators 
```suggestion
              && !infinite}
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,245 @@
...
+            <input
+              className="Carousel__form__input"
+              type="number"
+              value={step}
+              min="1"
+              max="5"
+              step="1"
+              onChange={(e) => {
+                this.setState({ step: +e.target.value });
+              }}
```
Anastasiia-Svintsova, line 182: Would be better to create a method for handling state changes

Anastasiia-Svintsova (approved):

src/components/Carousel.tsx
```diff
@@ -1,18 +1,268 @@
...
+        break;
+      case 'frameSize':
+        this.setState({ frameSize: +value });
+        break;
+      case 'imageSize':
+        this.setState({ frameSize: +value });
+        break;
+      case 'moutionSpeed':
+        this.setState({ frameSize: +value });
+        break;
```
anstsot, line 122: you're updating here the same values for different inputs. please check it because the demo seems to be broken (check video)

anstsot (changes requested):
The demo seems to work not as expected. could you fix it please?

https://user-images.githubusercontent.com/22824431/221825416-689f11fc-ebfb-4ec4-a047-1db24b0331e8.mov

src/components/Carousel.tsx
dima-korovai: you are right. i did changes and didn't check the final result

witflash (approved):
Well done 👍

PR #317, opened 2023-02-27
working slider, but tests 4  faild

webdevw:
[DEMO LINK](https://webdevw.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,22 @@
 import React from 'react';
 import './App.scss';
-import Carousel from './components/Carousel';
+import { Carousel } from './components/Carousel';
 
 interface State {
   images: string[];
+  itemWidth: number;
+  frameSize: number;
+  step: number;
+  animationDuration: number;
+  infinite: boolean;
 }
+
+type PropCarousel = (
+  'itemWidth'
+  | 'frameSize'
+  | 'step'
+  | 'animationDuration'
+);
 
 class App extends React.Component<{}, State> {
   state = {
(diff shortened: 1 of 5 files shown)


polosanya (changes requested):
image
Try to use these ids for your labels to pass the tests

src/App.tsx
```diff
@@ -20,17 +32,133 @@ class App extends React.Component<{}, State> {
...
+          infinite={infinite}
+        />
+
+        <div className="form">
+          <label htmlFor="itemId">
+            <input
+              type="number"
+              min={0}
+              id="itemId"
+              defaultValue={itemWidth}
```
agniya-i, line 102: try to change `defaultValue` to `value` and check if the test passes 🤔

src/App.tsx
```diff
@@ -20,17 +32,133 @@ class App extends React.Component<{}, State> {
...
+
+          <label htmlFor="stepId">
+            <input
+              defaultValue={step}
+              type="number"
+              min={1}
+              id="stepId"
+              onChange={(event) => (
+                this.handlePropsCarousel(
+                  'step',
```
agniya-i, line 137: you can also use `event.target.name`

src/App.tsx
```diff
@@ -20,17 +32,133 @@ class App extends React.Component<{}, State> {
...
     ],
+    itemWidth: 130,
+    frameSize: 3,
+    step: 3,
+    animationDuration: 1000,
+    infinite: false,
+  };
+
+  handlePropsCarousel = (property: PropCarousel, value: number) => {
+    switch (property) {
```
agniya-i, line 43: do we need this switch case? 
we can set value to state in this way `setState({[property] : value })`

src/components/Carousel.tsx
```diff
@@ -1,18 +1,194 @@
...
+                step,
+                itemWidth,
+                'next',
+                images.length,
+                frameSize,
+                infinite,
+              );
+            }}
+            disabled={
+              (itemShift === images.length - frameSize) && true
```
agniya-i, line 185: `&& true` looks redutant

src/components/Carousel.tsx
```diff
@@ -1,18 +1,194 @@
...
+              right: `${offset}px`,
+              transitionDuration: `${animationDuration}ms`,
+            }
+          }
+        >
+          {images.map((image, index) => (
+            <li key={image}>
+              <img
+                src={image}
+                alt={(index + 1).toString()}
```
agniya-i, line 143: better to give more details in alt, for example 
```suggestion
                alt={`image ${index+1}`}
```

agniya-i (changes requested):

src/App.tsx
webdevw: Changing to a value did not help to pass the tests

agniya-i (approved):

PR #319, opened 2023-02-28
solved

Deniso4ka23:
[DEMO LINK](https://Deniso4ka23.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 import React from 'react';
 import './App.scss';
 import './components/Form.scss';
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -20,17 +27,100 @@ class App extends React.Component<{}, State> {
...
+    this.setState({ frameSize: +newValue });
+  };
+
+  setItemWidth = (newValue: string) => {
+    this.setState({ itemWidth: +newValue });
+  };
+
+  setAnimationDuration = (newValue: string) => {
+    this.setState({ animationDuration: +newValue });
+  };
```
agniya-i, line 51: it would be better to create one common function, for example `handleChange` and pass name and value

agniya-i (changes requested):

PR #327, opened 2023-03-10
solution

ivanm2706:
[DEMO LINK](https://ivanm2706.github.io/react_carousel/)


YegorKorenkov (changes requested):
1. Please add default values to form controls
image
2. I can provide neganive values for frame size and step, please fix
image

VitaliyBondarenko1982 (approved):

PR #335, opened 2023-03-16
Develop

Vlad-Fedorishchev:
[DEMO LINK](https://Vlad-Fedorishchev.github.io/react_carousel/)


src/components/Field/Field.tsx
```diff
@@ -0,0 +1,132 @@
...
+  state = {
+    step: this.props.step,
+    frameSize: this.props.frameSize,
+    imageSize: this.props.imageSize,
+    animationDuration: this.props.animationDuration,
+    onChangeStep: this.props.onChangeStep,
+    onChangeFrameSize: this.props.onChangeFrameSize,
+    onChangeImageSize: this.props.onChangeImageSize,
+    onChangeAnimDuration: this.props.onChangeAnimDuration,
+  };
```
DarkMistyRoom, line 25: it is a bad idea to make a state based on props. you need to have only one source of trues

DarkMistyRoom (changes requested):
Good job, just improve it a bit:
- it looks like wants to be improved

image

- I don't understand the `step `limiting logic. why its value can't be more the frame size
image

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,139 @@
...
+          </button>
+
+          <button
+            data-cy="next"
+            type="button"
+            className="Carousel__nextButton Carousel__button"
+            disabled={nextButtonDis}
+            onClick={this.nextButton}
+          >
+            {}
```
witflash, line 130: You can use self-closed `button` tag instead of pass an empty block inside.

witflash (approved):
Good work 👍🏻

PR #343, opened 2023-03-24
Solution

dissolveyo:
[DEMO LINK](https://dissolveyo.github.io/react_carousel/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -3,11 +3,3 @@
   flex-direction: column;
   align-items: center;
 }
-
-label {
-  font-size: 20px;
-}
-
-input {
-  margin-top: 20px;
-}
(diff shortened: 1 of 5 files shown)


Volodymyr-Melnyk-Fron10der (changes requested):
keep it up)

1. next button should work in infinity mode
image

src/App.scss
```diff
@@ -1 +1,13 @@
-// style go here
+.App {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
+
+label {
```
OleziO, line 7: Try to use class selectors instead of tag names.

OleziO (changes requested):
image

Align center
image

etojeDenys (approved):
good job

PR #345, opened 2023-03-26
React Carousel

sergey-vorobei:
[DEMO LINK](https://sergey-vorobei.github.io/react_carousel/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@
 ## REQUIREMENTS:
 
 1. The title of the page should contain "Carousel"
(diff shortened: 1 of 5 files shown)


sTorba24 (changes requested):
It almost good!
make the carousel scroll, now the elements just disappear and new ones appear
And it should be like a real scroll, I should see part of the previous emoji while it scrolls|
try with styling `transform` and `transition` for ul

sTorba24 (approved):
Good work!

PR #350, opened 2023-04-04
Develop

RomeoKs:
[DEMO LINK](https://RomeoKs.github.io/react_carousel/)

### DIFF
--- a/src/components/InputRadio/InputRadio.scss
+++ b/src/components/InputRadio/InputRadio.scss
@@ -1,0 +1,16 @@
+.inputs {
+  &__container {
+    display: flex;
+    flex-direction: column;
+
+    &-radio {
+      display: grid;
+      grid-template-columns: 5fr 1fr 1fr 2fr;
+
+      label {
+        cursor: pointer;
+        display: flex;
+      }
+    }
+  }
+}
--- a/src/components/InputRange/InputRange.scss
+++ b/src/components/InputRange/InputRange.scss
@@ -1,0 +1,77 @@
+.inputs {
+  &__container {
+    display: flex;
+    flex-direction: column;
+
+    input {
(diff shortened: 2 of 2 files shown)


src/App.tsx
```diff
@@ -1,39 +1,29 @@
...
-      './img/7.png',
-      './img/8.png',
-      './img/9.png',
-      './img/10.png',
-    ],
-  };
+  return (
+    <div className="App">
+      {/* eslint-disable-next-line */}
+      <h1 className="App__title" data-cy="title">Carousel with {images.length} images</h1>
```
meforoma, line 22: not sure what linter complained about, but i'd multi-line it
```suggestion
      <h1 className="App__title" data-cy="title">
        Carousel with {images.length} images
      </h1>
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,323 @@
...
+    setFrameWidth(itemNumber * itemWidth);
+    setPosition(0);
+  }, [itemNumber, itemWidth]);
+
+  useEffect(() => {
+    if (infiniteRotation) {
+      setPosition(0);
+    }
+  }, [infiniteRotation]);
+
```
meforoma, line 46: i think you're missing another useEffect

the usecase it covers is: when you have 5 items in frame && step is 4
and you change items in frame to something that's below step, step is not re-applied -> images pass-by.

Try it on your current demo.

```suggestion

  useEffect(() => {
    if (itemNumber < step) {
      setSteps(itemNumber)
    }
  }, [itemNumber, step])
```

RomeoKs: I've looked into the issue and did not see any difference.
Need a second pair of goggles 🥽 😄

PR #351, opened 2023-04-04
Develop

lukacsja:
[DEMO LINK](https://lukacsja.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -15,15 +15,14 @@ const Carousel: React.FC<Props> = ({ images }) => {
   const containerWidth = (currentIndex + visibleImages) * imageSize;
 
   const handleNextClick = () => {
-    const nextIndex = currentIndex + step;
+    const imagesLeft = images.length - currentIndex - visibleImages;
+    const nextIndex = currentIndex + Math.min(step, imagesLeft);
 
-    if (nextIndex < images.length) {
-      setCurrentIndex(nextIndex);
-    }
+    setCurrentIndex(nextIndex);
   };
 
   const handlePrevClick = () => {
-    const prevIndex = currentIndex - step;
+    const prevIndex = currentIndex - Math.min(step, currentIndex);
 
     if (prevIndex >= 0) {
       setCurrentIndex(prevIndex);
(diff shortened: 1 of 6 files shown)


src/components/Carousel.tsx
```diff
@@ -11,8 +12,21 @@ const Carousel: React.FC<Props> = ({ images }) => {
...
-  const containerWidth = (currentIndex + visibleImages) * imageSize;
+  const wrapper = useCallback(
+    debounce((value: number) => {
+      setImageSize(value);
+    }, 1000),
+    [],
+  );
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const value = +event.target.value;
```
meforoma, line 25: better use explicit type conversion
```suggestion
const value = Number(event.target.value);

```

src/components/Carousel.tsx
```diff
@@ -29,11 +43,20 @@ const Carousel: React.FC<Props> = ({ images }) => {
     }
   };
 
+  const containerWidth = visibleImages * inputValue;
+
   const translateValue = `translateX(-${currentIndex * imageSize}px)`;
 
+  const isPrevDisabled = currentIndex === 0;
+  const isNextDisabled = currentIndex + visibleImages === images.length;
```
meforoma, line 51: it might happen to be >, so 
```suggestion
const isNextDisabled = currentIndex + visibleImages >= images.length;
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,154 @@
...
+
+  const containerWidth = visibleImages * inputValue;
+
+  const translateValue = `translateX(-${currentIndex * imageSize}px)`;
+
+  const isPrevDisabled = currentIndex === 0;
+  const isNextDisabled = currentIndex + visibleImages === images.length;
+
+  // STYLES >>>
+  const containerStylesOuter = {
```
meforoma, line 54: you could type-safe css props with respective type
```suggestion
  const containerStylesOuter: React.CSSProperties = {
```

meforoma (approved):
see the few comments
good otherwise 👍🏼

PR #355, opened 2023-04-09
Develop

ladrom:
[Demo](https://ladrom.github.io/react_carousel/)


polosanya (changes requested):
Good job! Found a few problems that would be great to fix)
1. Limit the frame size so it's not possible to set the negative value or zero
image
2. Same with other values
image
3. Item width moves the images
carousel

witflash (approved):
Well done 👍

PR #362, opened 2023-04-16
The Carousel has been implemented

bojkovladislav:


SanyaBratashchuk (changes requested):
It is ok to use libraries in many cases. But this task is about implement it without extra libraries

src/App.scss
```diff
@@ -1 +1,49 @@
-// style go here
+@import './components/Carousel.scss';
+
+*,
+*::before,
+*::after {
+  padding: 0;
+  margin: 0;
```
DarkMistyRoom, line 7: Don't use * selector for zeroing out your margins or paddings. It's still inefficient for browser to read your web document

src/App.scss
```diff
@@ -1 +1,49 @@
...
+
+.App {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  gap: 50px;
+}
+
+img {
```
DarkMistyRoom, line 23: it is better to use class

src/App.tsx
```diff
@@ -20,17 +25,117 @@ class App extends React.Component<{}, State> {
...
+          value={value.toString()}
+          min={min}
+          max={max}
+          onChange={(e) => {
+            const parsedValue = typeof value === 'number'
+              ? +e.target.value
+              : e.target.checked;
+
+            this.updateState(name as keyof State, parsedValue);
+          }}
```
DarkMistyRoom, line 66: move it to a handler function

DarkMistyRoom (approved):
Good job!

PR #363, opened 2023-04-19
react_carousel

Oleksandr-Maslo:
https://oleksandr-maslo.github.io/react_carousel/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -28,9 +28,16 @@ class App extends React.Component<{}, State> {
     return (
       <div className="App">
         {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1 data-cy="title">Carousel with {images.length} images</h1>
 
-        <Carousel />
+        <Carousel
+          images={images}
+          step={3}
+          frameSize={3}
+          itemWidth={130}
+          animationDuration={1000}
+          infinite={false}
+        />
       </div>
     );
   }
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,116 @@
-import React from 'react';
+import { Component } from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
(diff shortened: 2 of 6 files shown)


sTorba24 (changes requested):
Good start, keep it up! Implement form to change parameters

Oleksandr-Maslo:
But i didn`t fix problem with tests. I can`t run them

src/components/Carousel.tsx
```diff
@@ -1,18 +1,116 @@
...
+                { position: position + widthStep },
+              );
+            }
+
+            if (position === 0 && !infinite) {
+              this.setState(
+                { position: maxPosition },
+              );
+            }
+          }}
```
AlisaVlasova, line 83: move it to a separate function

AlisaVlasova (changes requested):
add default values and disable the possibility to enter negative values
image
add labels to form fields
check the demo to be sure that all fields work correctly

src/App.tsx
```diff
@@ -3,35 +3,154 @@ import './App.scss';
...
+  stateUpdate = (name: string, value: number | string) => {
+    this.setState((state) => {
+      return { ...state, [name]: value };
+    });
+  };
+
+  removeNegatives = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === '-') {
+      event.preventDefault();
+    }
```
DarkMistyRoom, line 44: set limits to the fields instead of changing the value, because now it doesn't work
image

src/App.tsx
```diff
@@ -3,35 +3,154 @@ import './App.scss';
...
+      infinite,
+    } = this.state;
 
     return (
-      <div className="App">
-        {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+      <>
+        <div className="App">
+          {/* eslint-disable-next-line */}
```
DarkMistyRoom, line 59: don't disable the linter, fix the issue

DarkMistyRoom (changes requested):
- `Item width` changes the frame size but not the item

image

- `Infinite animation` doesn't work

anstsot (changes requested):
Good job, but there is one fix. 
The images seem to be cut if you click on next and then chang the image size.
image

Let's observe some props in `Carousel` component (use `getDerivedStateFromProps`) and recalculate `position` state variable if needed.

TarasHoliuk (approved):
GJ :+1:

PR #365, opened 2023-04-21
React Carousel v.0.2 - data-cy added

NedobezhkinOleksandr:
https://NedobezhkinOleksandr.github.io/react_carousel/


src/App.tsx
```diff
@@ -20,20 +26,109 @@ class App extends React.Component<{}, State> {
...
+
+          <input
+            type="number"
+            id="frameSize"
+            min="1"
+            max={images.length}
+            value={frameSize}
+            onChange={(event) => this.setState({
+              frameSize: +event.target.value,
+            })}
```
SanyaBratashchuk, line 87: to avoid duplication of onChange callbacks you pass to the inputs you can add input names the same as in the state and do something like this `handleOnChange: (e) => {this.setState({[e.target.name]: event.target.value})} `

SanyaBratashchuk (approved):

PR #367, opened 2023-04-26
added carousel

PavloReutskiy:
[DEMO LINK](https://PavloReutskiy.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,175 @@
-import React from 'react';
+import React, { ChangeEvent, useState } from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
-      <li><img src="./img/1.png" alt="3" /></li>
-      <li><img src="./img/1.png" alt="4" /></li>
-    </ul>
+interface Props {
+  images: string[];
+}
 
-    <button type="button">Prev</button>
-    <button type="button">Next</button>
-  </div>
-);
+const Carousel: React.FC<Props> = ({ images }) => {
+  const [itemWidth, setItemWidth] = useState(130);
+  const [frameSize, setFrameSize] = useState(3);
+  const [step, setStep] = useState(3);
+  const [animationDuration, setAnimationDuration] = useState(1000);
+  const [scrollSize, setScrollSize] = useState(0);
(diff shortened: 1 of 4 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,174 @@
...
+const Carousel: React.FC<Props> = ({ images }) => {
+  const [itemWidth, setItemWidth] = useState(130);
+  const [frameSize, setFrameSize] = useState(3);
+  const [step, setStep] = useState(3);
+  const [animationDuration, setAnimationDuration] = useState(1000);
+  const [scrollSize, setScrollSize] = useState(0);
+  const [isRightButtonDisabled, setIsRightButtonDisabled] = useState(false);
+  const [isLeftButtonDisabled, setIsLeftButtonDisabled] = useState(true);
+
+  const widthOfContainer = itemWidth * 10;
```
roshirin, line 17: I would use images.length instead of hardcoding actual number. In case we'll have another images array.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,174 @@
...
+
+  const handleAnimationDuration = (event: ChangeEvent<HTMLInputElement>) => {
```
roshirin, line 19: Here and in further cases you'll get "NaN" in every input field if you try to enter not a number symbol. I would use a simple validation or type="number" inputs.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,174 @@
...
+  const handleFrameSize = (event: ChangeEvent<HTMLInputElement>) => {
+    setFrameSize(+event.target.value);
+  };
+
+  const handleStepsQuantity = (event: ChangeEvent<HTMLInputElement>) => {
+    setStep(+event.target.value);
+  };
+
+  const handleScrollRight = () => {
+    setIsLeftButtonDisabled(false);
```
roshirin, line 36: This will not work nice in edge cases. Like if you choose 10 slides to show, slider doesn't go anywhere but buttons change their disabled state.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,174 @@
...
+            type="button"
+            className="Carousel__button"
+            data-cy="next"
+            onClick={handleScrollRight}
+            disabled={isRightButtonDisabled}
+          >
+            {'>'}
+          </button>
+        </div>
+
```
roshirin, line 115: I think it's better to place the form in a separate component.

roshirin (approved):
Good job, nice slider!

I've made some suggestions in the comments section.

src/components/Carousel.tsx
liliia-livandovskaya: agree, use images.length

PR #371, opened 2023-04-28
Done

Kocha1991:
[DEMO LINK](https://Kocha1991.github.io/react_carousel/)


DarkMistyRoom (changes requested):
- please add needed data-cy attributes

image

- your page should have the ability to change all these settings
image
like in this example 
image

tetiana-ivanchenko (changes requested):
item width should change width of each item, not container

SanyaBratashchuk (changes requested):
1. add min/max values 
image
2. frame size is not working
3. Check logic 
image

src/App.scss
```diff
@@ -1 +1,44 @@
...
+  &__title {
+    text-align: center;
+  }
+}
+
+.carusel-wrapper {
+  margin: 0 auto;
+  padding: 10px;
+  overflow: hidden;
+  border: 2px solid rgb(146, 115, 11);
```
OleziO, line 11: Try to use HEX instead of RGBA

src/App.tsx
```diff
@@ -20,17 +17,133 @@ class App extends React.Component<{}, State> {
...
-        {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
 
-        <Carousel />
+        <h1 className="App__title title">
+          Carousel with
+          {' '}
+          {images.length}
+          {' '}
+          images
```
OleziO, line 55: Replace with template string.

[More info.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

src/components/Carousel.scss
```diff
@@ -1 +1,33 @@
...
+  }
+
+  &__wrapper {
+    display: flex;
+    justify-content: center;
+  }
+
+  &__btn {
+    background-color: white;
+    border: 2px solid #555;
```
OleziO, line 16: Move all repeatable colors to the variables.

[More info.](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

src/components/Carousel.tsx
```diff
@@ -1,18 +1,70 @@
...
+  const prev = () => setCurrentIndex(prevIndex => (infinite
+    ? (prevIndex - step + images.length) % images.length
+    : Math.max(prevIndex - step, 0)));
+
+  return (
+    <div className="Carousel">
+      <ul className="Carousel__list">
+        {images.map((img) => (
+          <li style={{
+            transform: `translateX(${-currentIndex * itemWidth}px)`,
```
OleziO, line 36: Add key attribute.

OleziO (changes requested):
Fix infinity mode
image

src/components/Carousel.tsx
```diff
@@ -1,18 +1,88 @@
...
+
+      return Math.min(prev + step, images.length - frameSize);
+    }
+
+    return Math.min(prev + step, images.length - frameSize);
+  });
+
+  const prev = () => setCurrentIndex(prevIndex => {
+    if (infinite) {
+      if (prevIndex === 0) {
```
sTorba24, line 37:
```suggestion
      if (!prevIndex) {
```

sTorba24 (approved):
Well done!

PR #374, opened 2023-04-29
Develop

LBondarchuk:
[DEMO LINK](https://LBondarchuk.github.io/react_carousel/)


src/App.scss
```diff
@@ -1 +1,69 @@
...
+
+.form {
+  display: flex;
+  flex-direction: column;
+  border: 2px solid black;
+  border-radius: 10%;
+  width: max-content;
+  margin: 0 auto;
+  padding: 10px;
+  background-color: burlywood;
```
OleziO, line 21: Try to use HEX instead of color names.

src/App.tsx
```diff
@@ -20,17 +25,129 @@ class App extends React.Component<{}, State> {
...
-        <h1>Carousel with {images.length} images</h1>
+        <h1
+          className="App__title"
+          data-cy="title"
+        >
+          Carousel with
+          {' '}
+          {images.length}
+          {' '}
+          images
```
OleziO, line 56: Use template strings.

[More info.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

OleziO: Check all similar cases.

OleziO (changes requested):

Volodymyr-Melnyk-Fron10der (changes requested):
1. Infinite mode shouldn't trigger the carousel, you still need to click prev/next buttons, but it will be infinity, not blocked by sides
2. there is no animation, add smooth

anstsot (changes requested):
Good work! but the infinite state still isn't working. Please, fix it

anstsot (approved):
Good job!

LBondarchuk:
thank you

PR #380, opened 2023-05-02
Develop

nazar-medushevskyi:
- [DEMO LINK](https://nazar-medushevskyi.github.io/react_carousel/)


DarkMistyRoom (changes requested):
It looks good but according to the task, we should have a frame containing visible items - 3 by default. But now all items are visible)
image

IvanFesenko (approved):
Well done

PR #382, opened 2023-05-04
solution

nastychenko:
https://nastychenko.github.io/react_carousel/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,17 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
-import Carousel from './components/Carousel';
+import { Carousel } from './components/Carousel';
 
 interface State {
   images: string[];
+  step: number,
+  frameSize: number,
+  itemWidth: number,
+  animationDuration: number,
+  infinite: boolean,
 }
 
-class App extends React.Component<{}, State> {
+export class App extends Component<{}, State> {
   state = {
     images: [
       './img/1.png',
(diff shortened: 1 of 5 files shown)


src/components/Carousel.scss
```diff
@@ -1 +1,21 @@
-// style go here
+.carousel {
+  overflow: hidden;
+
+  &__list {
+    display: flex;
+    list-style: none;
+    padding: 0;
+    margin: 0;
+    width: 1300px;
```
polosanya, line 9: It would be great to use variables for main values

src/components/Carousel.tsx
```diff
@@ -1,18 +1,152 @@
-import React from 'react';
+/* eslint-disable react/no-unused-prop-types */
```
polosanya, line 1: Let's try fix eslint problem instead of disabling it

src/components/Carousel.tsx
```diff
@@ -1,18 +1,152 @@
...
+      if (currentSlide + itemWidth * step > 0) {
+        this.setState({ currentSlide: 0 });
+      } else {
+        this.setState({ currentSlide: currentSlide + itemWidth * step });
+      }
+    } else if (infinite && !currentSlide) {
+      this.setState({ currentSlide: -maxSlide });
+    } else {
+      this.setState({ currentSlide: currentSlide - itemWidth * step });
+    }
```
polosanya, line 49: When using 'setState', it is asynchronous. So, when you set the new state and try to access it right away, you might get the old state. To fix this, use the setState callback function to perform actions after the state is set.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,152 @@
...
+          ref={this.listRef}
+          className="carousel__list"
+          style={{
+            transform: `translateX(${currentSlide}px)`,
+            transition: `all ${animationDuration}ms ease-in-out`,
+            width: `${itemWidth * images.length}px`,
+          }}
+        >
+          {images.map(img => (
+            <li key={images.indexOf(img)}>
```
polosanya, line 122: Never use index as a key
https://www.developerway.com/posts/react-key-attribute

polosanya (changes requested):
Good job! Let's try to improve your solution a bit)

src/App.tsx
```diff
@@ -20,20 +25,99 @@ class App extends React.Component<{}, State> {
...
+    itemWidth: 130,
+    animationDuration: 1000,
+    infinite: false,
   };
 
   render() {
-    const { images } = this.state;
+    const {
+      images, step, frameSize, itemWidth, animationDuration, infinite,
+    } = this.state;
```
loralevitska, line 38:
```suggestion
    const {
      images,
      step,
      frameSize,
      itemWidth,
      animationDuration,
      infinite,
    } = this.state;
```

src/App.tsx
```diff
@@ -20,20 +25,99 @@ class App extends React.Component<{}, State> {
...
+            Item width
+            <br />
+            <input
+              type="number"
+              id="itemWidth"
+              defaultValue={130}
+              min={130}
+              onChange={(e) => {
+                this.setState({ itemWidth: +e.target.value });
+              }}
```
loralevitska, line 65: Move to separate function  (transfer the function to a variable, and here only write the name of the function). Fix all cases if present

src/components/Carousel.tsx
```diff
@@ -1,18 +1,153 @@
...
+  images: string[];
+  step: number,
+  frameSize: number,
+  itemWidth: number,
+  animationDuration: number,
+  infinite: boolean,
+}
+
+export class Carousel extends React.Component<Props, State> {
+  private listRef = createRef<HTMLUListElement>(); // create ref for carousel__list
```
loralevitska, line 18: Remove comment. Try it write obvious code that explain all we need without comments)

loralevitska (changes requested):
Almost done)
Let's improve

polosanya (approved):

PR #387, opened 2023-05-08
Solution

8Melya8:
[DEMO LINK](https://8melya8.github.io/react_carousel/)


src/App.scss
```diff
@@ -1 +1,34 @@
-// style go here
```
etojeDenys, line 1: Use block comments instead of line comments for the file header.

etojeDenys (approved):
LGTM

PR #393, opened 2023-05-20
Develop

dianaavem:
[DEMO LINK](https://dianaavem.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,7 +3,12 @@ import './App.scss';
 import Carousel from './components/Carousel';
 
 interface State {
-  images: string[];
+  images: string[],
+  step: number,
+  frameSize: number,
+  itemWidth: number,
+  animationDuration: number,
+  infinite: boolean,
 }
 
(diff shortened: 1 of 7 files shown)


SanyaBratashchuk (changes requested):
Animation is not working

SanyaBratashchuk (changes requested):
There is shouldn't an empty space 
image

src/App.tsx
```diff
@@ -20,17 +25,135 @@ class App extends React.Component<{}, State> {
...
+              id="step"
+              type="text"
+              className="field__input"
+              value={step}
+              min={1}
+              max={images.length}
+              onChange={(e) => {
+                this.setState({
+                  step: +e.target.value,
+                });
```
SanyaBratashchuk, line 75: create handle change common  for all inputs. It may work like that `handleChange: (e) => {this.setState({[e.target.name]: e.target.value})}`

SanyaBratashchuk (approved):

PR #398, opened 2023-05-24
Develop

Zykhor-Andrii:
[TASK LINK](https://Zykhor-Andrii.github.io/react_carousel/)


src/App.scss
```diff
@@ -1 +1,24 @@
 // style go here
```
OleziO, line 1: Remove all comments from the code.

PR #406, opened 2023-06-08
Solution

Ukrainiane-panda:
[DEMO LINK](https://Ukrainiane-panda.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,6 +4,11 @@ import Carousel from './components/Carousel';
 
 interface State {
   images: string[];
+  itemWidth: number,
+  frameSize: number,
+  step: number,
+  animationDuration: number,
+  infinity: boolean,
 }
 
 class App extends React.Component<{}, State> {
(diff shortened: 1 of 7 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,129 @@
...
+            onClick={() => this.handlePrev(-step)}
+          >
+            Prev
+          </button>
+          <button
+            type="button"
+            data-cy="next"
+            className="button__nav"
+            disabled={nextDisabled}
+            onClick={() => this.handleNext(step)}
```
TarasHoliuk, line 119:
```suggestion
            onClick={this.handleNext}
```
you may get `step` from props inside a function, in the same way as other values (e.g. `frameSize`)

src/components/Carousel.tsx
```diff
@@ -1,18 +1,129 @@
...
+            })
+          }
+        </ul>
+
+        <div className="button">
+          <button
+            type="button"
+            className="button__nav"
+            disabled={prevDisabled}
+            onClick={() => this.handlePrev(-step)}
```
TarasHoliuk, line 110: the same as below, handle negative value inside `handlePrev` function

src/App.tsx
```diff
@@ -20,17 +25,128 @@ class App extends React.Component<{}, State> {
...
+
+  setFrameSize = (event: React.ChangeEvent<HTMLInputElement>) => {
+    this.setState({ frameSize: +event.target.value });
+  };
+
+  setStep = (event: React.ChangeEvent<HTMLInputElement>) => {
+    this.setState({ step: +event.target.value });
+  };
+
+  setAnimationDuration = (event: React.ChangeEvent<HTMLInputElement>) => {
```
TarasHoliuk, line 47: you may unify all functions with `number` value - pass `name` prop to each input that matches key in the state, then use it in the handler (event.target.name) to set proper value in the state

TarasHoliuk (changes requested):
very good job :+1: 
I've added a few suggestions to make it perfect

volodymyr-soltys97 (changes requested):
test

volodymyr-soltys97 (approved):

PR #409, opened 2023-06-10
infinite Carousel

victor-buldenko:
[DEMO LINK](https://victor-buldenko.github.io/react_carousel/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,8 @@
-// style go here
+* {
+  margin: 0;
+  padding: 0;
+}
+
+body {
+  background-color:rgb(86, 134, 134);
+}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,16 @@
+/* eslint-disable no-console */
+/* eslint-disable padded-blocks */
 import React from 'react';
 import './App.scss';
-import Carousel from './components/Carousel';
+import { Carousel } from './components/Carousel';
 
 interface State {
   images: string[];
+  step: number,
+  frameSize: number;
+  itemWidth: number;
+  animationDuration: number;
+  infinite: boolean,
 }
 
 class App extends React.Component<{}, State> {
@@ -20,17 +27,146 @@ class App extends React.Component<{}, State> {
       './img/9.png',
       './img/10.png',
     ],
+    step: 3,
+    frameSize: 3,
+    itemWidth: 130,
+    animationDuration: 1000,
+    infinite: false,
+  };
+
+  inputHandler = (e: { currentTarget: HTMLInputElement }) => {
+    const { name, value } = e.currentTarget;
+
+    switch (name) {
+      case 'step':
+        this.setState({ step: +value });
+        break;
+
+      case 'frameSize':
+        this.setState({ frameSize: +value });
+        break;
+
+      case 'itemWidth':
+        this.setState({ itemWidth: +value });
+        break;
+
+      case 'animationDuration':
+        this.setState({ animationDuration: +value });
+        break;
+
+      default:
+        this.setState({ infinite: e.currentTarget.checked });
+        break;
+    }
   };
 
   render() {
-    const { images } = this.state;
+    const {
+      images,
+      step = 3,
+      frameSize = 3,
+      itemWidth = 130,
+      animationDuration = 1000,
+      infinite = false,
+    } = this.state;
 
     return (
       <div className="App">
         {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1
+          data-cy="title"
+          className="title"
+        >
+          Carousel with
+          {images.length}
+          {' '}
+          images
+        </h1>
 
-        <Carousel />
+        <Carousel
+          images={images}
+          step={step}
+          frameSize={frameSize}
+          itemWidth={itemWidth}
+          animationDuration={animationDuration}
+          infinite={infinite}
+          onAutoplayChange={(autoplay) => {
+            this.setState({ infinite: autoplay });
+          }}
+        />
+
+        <form action="#" className="form">
+          <label className="label">
+            slide step
+            {` [${step}]`}
+            <input
+              className="input"
+              type="number"
+              name="step"
+              id="inputStep"
+              min="1"
+              max="3"
+              placeholder="1-3"
+              onChange={this.inputHandler}
+            />
+          </label>
+          <label className="label">
+            frame size
+            {` [${frameSize}]`}
+            <input
+              className="input"
+              type="number"
+              name="frameSize"
+              id="frameSize"
+              min="1"
+              max="5"
+              placeholder="1-5"
+              onChange={this.inputHandler}
+            />
+          </label>
+          <label className="label">
+            slide width
+            {` [${itemWidth}]`}
+            <input
+              className="input"
+              type="number"
+              name="itemWidth"
+              id="itemWidth"
+              min="130"
+              max="350"
+              step="20"
+              placeholder="100-350"
+              onChange={this.inputHandler}
+            />
+          </label>
+          <label className="label">
+            animation(ms)
+            {` [${animationDuration}]`}
+            <input
+              className="input"
+              type="number"
+              name="animationDuration"
+              id="animationDuration"
+              min="100"
+              max="5000"
+              step="100"
+              placeholder="100-5000"
+              onChange={this.inputHandler}
+            />
+          </label>
+          <label className="label">
+            autoplay
+            {` [${this.state.infinite}]`}
+            <input
+              className="checkbox"
+              type="checkbox"
+              name="infinite"
+              id="infinite"
+              checked={this.state.infinite}
+              onChange={this.inputHandler}
+            />
+          </label>
+        </form>
       </div>
     );
   }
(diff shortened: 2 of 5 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,224 @@
...
+
+    if (this.props.onAutoplayChange) {
+      this.props.onAutoplayChange(false);
+    }
+  };
+
+  slidePrev = () => {
+    clearInterval(this.autoplayId);
+    const buttons = Array.from(
+      document.querySelectorAll('.button'),
```
TarasHoliuk, line 86: that's not the way (using `querySelector` as in the vanilla JS) we should work in React (see comment below) - you have access to elements in JSX below (in `render` method)

if you need DOM element you should use `ref` (but you don't in this case)

src/components/Carousel.tsx
```diff
@@ -1,18 +1,224 @@
...
+    ) as HTMLButtonElement;
+    const buttonNext = document.querySelector(
+      '.buttonNext',
+    ) as HTMLButtonElement;
+
+    buttonNext.removeEventListener('click', this.slideNext);
+    buttonPrev.removeEventListener('click', this.slidePrev);
+  }
+
+  autoplay = (): number => {
```
TarasHoliuk, line 55: that looks very nice :+1: 
but infinity should do a bit different thing

let's keep autoplay under a different name and implement `infinity` as described in the task requirements

src/components/Carousel.tsx
```diff
@@ -1,18 +1,224 @@
...
+  onAutoplayChange: (autoplay: boolean) => any;
+};
+
+type State = {
+  images: string[];
+};
+
+export class Carousel extends React.Component<Props, State> {
+  state = {
+    images: this.props.images,
```
TarasHoliuk, line 21: no need to save images in the state and manipulate them
store the index of the item you want to display in the first place
then, on the buttons click recalculate the expected index

use it to `transform: translate` (it's better for performance than `margin`) `li` item

TarasHoliuk (changes requested):

src/components/Carousel.tsx
victor-buldenko: rejected, because the carousel is a circle by definition - its means infinity. Counter for listed slides make yourself if you need.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,224 @@
...
-export default Carousel;
+type Props = {
+  images: string[];
+  step: number;
+  frameSize: number;
+  itemWidth: number;
+  animationDuration: number;
+  infinite: boolean;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  onAutoplayChange: (autoplay: boolean) => any;
```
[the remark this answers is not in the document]
victor-buldenko: there is a special type of 'any', if someone wants to change the callback, he won't have any problems and won't have to write extra checks, it doesn't return anything anyway, so there is no sense in 'void'.

src/App.tsx
```diff
@@ -20,17 +26,148 @@ class App extends React.Component<{}, State> {
...
+    animationDuration: 1000,
+    infinite: false,
+    autoplay: false,
+  };
+
+  inputHandler = (e: { currentTarget: HTMLInputElement }) => {
+    const { name, value } = e.currentTarget;
+
+    switch (name) {
+      case 'step':
```
OleziO, line 41: Move all cases to enum.

src/components/Carousel.scss
```diff
@@ -1 +1,108 @@
...
+
+    img {
+      will-change: transform;
+      width: 100%;
+      height: 100%;
+    }
+  }
+
+  .button {
+    border: 1px solid #fdfffc;
```
OleziO, line 41: Move all repeatable colors to variables.

src/components/Carousel.scss
```diff
@@ -1 +1,108 @@
...
+  padding-left: 10px;
+  align-items: baseline;
+  gap: 10px;
+}
+
+.input {
+  outline: none;
+  border: none;
+  width: 80px;
+  font-family: "Montserrat", sans-serif;
```
OleziO, line 96: Move to html selector and remove from all other places.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,299 @@
...
+      slides.length - this.props.step,
+    );
+
+    urls.unshift(...lastUrls);
+    slides.unshift(...LastSlides);
+    this.setState({ images: urls });
+
+    const timerId = setTimeout(() => {
+      clearInterval(timerId);
+      for (let i = 0; i < images.length; i += 1) {
```
OleziO, line 150: Replace `i += 1` with `i++`. Check similar cases.

OleziO (changes requested):

src/components/Carousel.tsx
victor-buldenko: image
next time, maybe, before rejecting the task, would you be sure you are right?

src/App.tsx
```diff
@@ -20,17 +26,157 @@ class App extends React.Component<{}, State> {
...
+  };
+
+  inputHandler = (e: { currentTarget: HTMLInputElement }) => {
+    enum EnumInputProp {
+      Step = 'step',
+      FrameSize = 'frameSize',
+      ItemWidth = 'itemWidth',
+      AnimationDuration = 'animationDuration',
+      Autoplay = 'autoplay',
+    }
```
etojeDenys, line 44: let's move this enum out of the component

src/App.tsx
```diff
@@ -20,17 +26,157 @@ class App extends React.Component<{}, State> {
...
+          { ...prevState, [name]: +value }));
+        break;
+
+      case EnumInputProp.Autoplay:
+        this.setState({ autoplay: e.currentTarget.checked, infinite: false });
+        break;
+
+      default:
+        this.setState({ infinite: e.currentTarget.checked, autoplay: false });
+        break;
```
etojeDenys, line 63:
```suggestion
      case EnumInputProp.Autoplay:
        default:
        this.setState({ autoplay: e.currentTarget.checked, infinite: false });
        break;
```

etojeDenys (changes requested):

src/App.tsx
victor-buldenko: it's class method...

src/components/Carousel.tsx
DarkMistyRoom: Don't save props to the state. We need to have only one source of truth

DarkMistyRoom (changes requested):
Great job! Only one comment

src/App.scss
```diff
@@ -1 +1,14 @@
...
+  font-family: "Montserrat", sans-serif;
+}
+
+* {
+  margin: 0;
+  padding: 0;
+}
+
+body {
+  background-color: teal;
```
OleziO, line 13: Replace with HEX color code.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,301 @@
...
+};
+
+type State = {
+  images: string[];
+  disabled : boolean;
+};
+
+export class Carousel extends React.Component<Props, State> {
+  autoplayId = 0;
+
```
OleziO, line 22: Remove the empty line.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,301 @@
...
+  counter = 0;
+
```
OleziO, line 24: Remove the empty line.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,301 @@
...
+  showLastSlide = false;
+
```
OleziO, line 26: Remove the empty line.

OleziO (changes requested):

src/components/Carousel.tsx
victor-buldenko: image

src/components/Carousel.tsx
```diff
@@ -1,18 +1,301 @@
...
+      }
+    }, 0);
+
+    const timerId2 = setTimeout(() => {
+      clearInterval(timerId2);
+      this.setState({ disabled: false });
+    }, this.props.animationDuration);
+  };
+
+  slideNext = () => {
```
loralevitska, line 164: If your method or function is larger than 2 laptop viewports, consider using helper functions and avoid large code nesting (more than 3) as this makes your code harder to read and maintain for other developers on your team.

image

loralevitska (changes requested):
Good job! 
You can improve your code. Check my comments:

- check the conditions under which the code works. You are currently working in such a way that when the infinitive checkbox is not enabled, the carousel works exactly in the infinitive, and then vice versa

src/components/Carousel.tsx
```diff
@@ -1,18 +1,301 @@
...
+export class Carousel extends React.Component<Props, State> {
+  autoplayId = 0;
+
+  counter = 0;
+
+  showLastSlide = false;
+
+  imagesSrc = this.props.images;
+
+  state = {
```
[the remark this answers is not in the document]
victor-buldenko: image
are all of you blind?

loralevitska (approved):

PR #410, opened 2023-06-11
react_carousel

YelyzavetaMolochynska:
[DEMO LINK](https://YelyzavetaMolochynska.github.io/react_carousel/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,25 @@
-// style go here
+.App {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  margin: 0 auto;
+}
+
+.inputs-container {
+  display: flex;
+  flex-direction: column;
+  gap: 5px;
+}
+
+input {
+  box-sizing: border-box;
+  width: 70px;
+  height: fit-content;
+  border: none;
+  border-bottom: 0.5px solid #000;
+  padding-left: 10px;
+}
+
+label {
+  font: 16px sans-serif;
+}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,14 @@
 import React from 'react';
 import './App.scss';
-import Carousel from './components/Carousel';
+import { Carousel } from './components/Carousel';
 
 interface State {
   images: string[];
+  step: number,
+  frameSize: number,
+  imgWidth: number,
+  animationDur: number,
+  infinite: boolean,
 }
 
 class App extends React.Component<{}, State> {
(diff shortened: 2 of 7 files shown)


src/App.scss
```diff
@@ -1 +1,25 @@
-// style go here
```
etojeDenys, line 1: Use block comments instead of line comments. Replace '//' with '/* */' for proper CSS commenting.

src/App.tsx
```diff
@@ -20,17 +25,127 @@ class App extends React.Component<{}, State> {
...
+
+        return;
+      }
+
+      case 'animationDuration': {
+        if (+event.currentTarget.value > 500) {
+          this.setState({ animationDur: +event.currentTarget.value });
+        }
+
+        this.setState({ animationDur: 1000 });
```
etojeDenys, line 74: Set the default animationDur only if the input value is not valid. Move this line inside the if statement's else block.

src/App.tsx
```diff
@@ -20,17 +25,127 @@ class App extends React.Component<{}, State> {
...
+        return;
+      }
+
+      case 'frameSize': {
+        if (+event.currentTarget.value
+          && +event.currentTarget.value <= this.state.images.length) {
+          this.setState({ frameSize: +event.currentTarget.value });
+        }
+
+        this.setState({ frameSize: 3 });
```
etojeDenys, line 54: Set the default frameSize only if the input value is not valid. Move this line inside the if statement's else block.

src/App.tsx
```diff
@@ -20,17 +25,127 @@ class App extends React.Component<{}, State> {
...
+
+        return;
+      }
+
+      case 'step': {
+        if (+event.currentTarget.value > 0) {
+          this.setState({ step: +event.currentTarget.value });
+        }
+
+        this.setState({ step: 3 });
```
etojeDenys, line 64: Set the default step only if the input value is not valid. Move this line inside the if statement's else block.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,135 @@
...
+          }}
+        >
+          {images.map((image, index) => (
+            <li
+              key={`${index + 1}`}
+              style={{ width: `${itemWidth}px`, height: `${itemWidth}px` }}
+            >
+              <img
+                src={image}
+                alt="1"
```
etojeDenys, line 108: The 'alt' attribute on the 'img' element should be dynamic, e.g., 'alt={`Image ${index + 1}`}'. This provides a better description for screen readers and improves accessibility.

etojeDenys (changes requested):
image
it does not work

src/App.tsx
```diff
@@ -20,17 +25,127 @@ class App extends React.Component<{}, State> {
...
+        <div className="inputs-container">
+          <label htmlFor="itemWidth">
+            Image size:
+            <input
+              type="number"
+              name="itemWidth"
+              id="itemWidth"
+              min={50}
+              placeholder="130"
+              onBlur={this.inputHandler}
```
VitaliyBondarenko1982, line 113: state of inputs should be updates onChange event
```suggestion
              onChange={this.inputHandler}
```
carousel

VitaliyBondarenko1982 (changes requested):

demetrxx (approved):
Well done 👍 Go ahead!

PR #413, opened 2023-06-19
react_carousel

vintwp:
https://vintwp.github.io/react_carousel/


src/App.scss
```diff
@@ -1 +1,21 @@
-// style go here
+* {
+  box-sizing: border-box;
+}
+
+iframe {
```
volodymyr-soltys97, line 5: The 'iframe' selector is not needed for this task. Consider removing it.

src/components/CarouselList/CarouselList.scss
```diff
@@ -0,0 +1,29 @@
...
+    // flex-shrink: 0;
+    // flex-grow: 0;
+
+    // &::after {
+    //   content: "";
+    //   display: block;
+    //   padding: 0 0 100%;
+    // }
+
+    img {
```
volodymyr-soltys97, line 19: Uncomment the necessary styles for the img element inside the carousel item.

src/components/CarouselList/CarouselList.scss
```diff
@@ -0,0 +1,29 @@
...
+    display: flex;
+    list-style: none;
+  }
+
+  &__item {
+    // position: relative;
+    // flex-shrink: 0;
+    // flex-grow: 0;
+
+    // &::after {
```
volodymyr-soltys97, line 13: Uncomment the styles for the carousel item's ::after pseudo-element.

src/components/CarouselList/CarouselList.scss
```diff
@@ -0,0 +1,29 @@
+.Carousel {
+
+  &__list {
+    display: flex;
+    list-style: none;
+  }
+
+  &__item {
+    // position: relative;
```
volodymyr-soltys97, line 9: Uncomment the necessary styles for the carousel item.

src/components/Carousel/Carousel.scss
```diff
@@ -0,0 +1,42 @@
...
+  &__wrapper {
+    margin: 0 auto;
+    overflow: hidden;
+  }
+
+  &__buttons {
+    display: flex;
+    justify-content: space-between;
+    padding: 10px 10px 0;
+    background-color: rgb(198, 206, 206);
```
volodymyr-soltys97, line 17: Consider using a variable for the background-color to maintain consistency and make it easy to update in the future.

src/components/Carousel/Carousel.scss
```diff
@@ -0,0 +1,42 @@
+$border-radius: 10px;
+
+.Carousel {
+  margin: 0 auto;
+  background-color: rgb(19, 174, 235);
```
volodymyr-soltys97, line 5: Consider using a variable for the background-color to maintain consistency and make it easy to update in the future.

src/components/CarouselList/CarouselList.tsx
```diff
@@ -0,0 +1,44 @@
...
+        key={`${image}+${Math.random()}}`}
+        className="Carousel__item"
+        style={{
+          width: `${itemWidth}px`,
+        }}
+      >
+        <img
+          src={image}
+          alt={String(index + 1)}
+          width={itemWidth}
```
volodymyr-soltys97, line 39: The width prop in the img tag is not needed, since the width is already being set in the inline style of the li element.

src/components/CarouselList/CarouselList.tsx
```diff
@@ -0,0 +1,44 @@
...
+  <ul
+    className="Carousel__list"
+    style={{
+      transform: `${transform}`,
+      transition: `${transition}`,
+    }}
+  >
+    {images.map((image, index) => (
+      <li
+        key={`${image}+${Math.random()}}`}
```
volodymyr-soltys97, line 30: Using Math.random() in a key prop can lead to unexpected issues. Consider using a unique and deterministic value, like the image URL or the index.

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+      this.setState({
+        offset: currentOffset,
+      });
+    }
+  };
+
+  addCloneSlidesForInfinitySlider = () => {
+    const imagesFromServer = this.props.images;
+    const { step, itemWidth } = this.props;
```
volodymyr-soltys97, line 150: You can use destructuring assignment to make the code cleaner. For example: 'const { images, step, itemWidth } = this.props;'

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+    const imagesWithClones = [...this.props.images];
+
+    imagesWithClones.push(...imagesFromServer.slice(0, step));
+    imagesWithClones.unshift(...imagesFromServer.slice(-step));
```
volodymyr-soltys97, line 155: You can use the spread operator to make the code cleaner. For example: 'const imagesWithClones = [...imagesFromServer.slice(-step), ...images];'

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

src/components/CarouselList/CarouselList.tsx
vintwp: Is it correct use index for key (Linter shows error) ?

vlad-demchuk: it's not

you can use just image if it's unique
if it's not, let's leave it as it is temporarily

src/components/Carousel/Carousel.tsx
vlad-demchuk: I am not sure you can replace all methods with spread operator and sometimes it will decrease readability and so.
but with the case with push it suits

vlad-demchuk (changes requested):
infinite doesn't work
image

src/components/CarouselList/CarouselList.tsx
vintwp: > it's not
> 
> you can use just image if it's unique if it's not, let's leave it as it is temporarily

But, if I clone slides for correct work of inifite slider, can you advice me what shall i use in key. During this task, i've seen a bug of updating components, when keys was duplicated. Thanks !

volodymyr-soltys97 (approved):
Good job 👍

PR #420, opened 2023-06-26
Solution

bohdan-mykhailenko:
[DEMO LINK](https://bohdan-mykhailenko.github.io/react_carousel/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,9 @@
-// style go here
+.App {
+  text-align: center;
+  margin-top: 50px;
+
+  h1 {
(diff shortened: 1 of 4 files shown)


roman-mirzoian (changes requested):
Now I see that part of the work is done well, but unfortunately, many items from the task were not completed, such as, for example, the ability to set parameters through the application itself (and not manually), appearance, etc.
Also, now the application itself does not work correctly:
https://gyazo.com/fc0bd14d8779cb7223bd51a46173bdcd

PR #423, opened 2023-06-29
solution

MaxSchmide:
[DEMO LINK](https://maxschmide.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -60,41 +60,49 @@ class App extends React.Component<{}, State> {
         {/* eslint-disable-next-line */}
         <h1 data-cy='title' className='App__title'>Carousel with {images.length} images</h1>
         <form className="App__form">
-          <label htmlFor="animationDuration">Animation duration (ms)</label>
-          <input
-            type="number"
-            name="animationDuration"
-            id="animationDuration"
-            value={animationDuration}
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -20,17 +26,92 @@ class App extends React.Component<{}, State> {
...
+        <h1 data-cy='title' className='App__title'>Carousel with {images.length} images</h1>
+        <form className="App__form">
+          <div className="App__field">
+            <label htmlFor="animationDuration">Animation duration (ms)</label>
+            <input
+              type="number"
+              name="animationDuration"
+              id="animationDuration"
+              value={animationDuration}
+              onChange={(e) => this.handleInputChange(e)}
```
ilona-borkovska, line 70: This remark is connected with all your inputs: you should set limits for input values (-1 or 10000 are not valid nums for carousel)

ilona-borkovska (changes requested):
Good job! But I am not sure that we should do this task by using class components

And678 (changes requested):
Looks very nice, but tests need to pass https://github.com/mate-academy/react_carousel/pull/423/checks

And678 (approved):
Approve, discussed in slack

PR #426, opened 2023-07-02
add task solution

ilona-borkovska:
[DEMO LINK](https://ilona-borkovska.github.io/react_carousel/)


ollavka (approved):
Nice)

src/components/Settings/Settings.tsx
```diff
@@ -0,0 +1,101 @@
...
+  step: number,
+  setStep: (value: number) => void,
+  animationDuration: number,
+  setAnimationDuration: (value: number) => void,
+}
+
+export const Settings: React.FC<Props> = ({
+  itemWidth,
+  setItemWidth,
+  IMAGE_COUNT,
```
qoa11a, line 20:
```suggestion
  imageCount: number,
  frameSize: number,
  setFrameSize: (value: number) => void,
  step: number,
  setStep: (value: number) => void,
  animationDuration: number,
  setAnimationDuration: (value: number) => void,
}

export const Settings: React.FC<Props> = ({
  itemWidth,
  setItemWidth,
  imageCount,
```
Here it would be better to name it using camel case as here it's not a constant with only one possible value. It may be changed.

qoa11a (approved):
Nicely done! ✅

PR #439, opened 2023-07-10
FedenkoReactCarousel

VladFedenko:
https://VladFedenko.github.io/react_carousel/

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,192 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
-      <li><img src="./img/1.png" alt="3" /></li>
-      <li><img src="./img/1.png" alt="4" /></li>
-    </ul>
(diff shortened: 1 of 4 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,192 @@
...
+    setStepMove(+evnt.target.value);
+  }
+
+  function changeAnimation(evnt: React.ChangeEvent<HTMLInputElement>) {
+    setAnimatonMove(+evnt.target.value);
+  }
+
+  function changeInfinite(evnt: React.ChangeEvent<HTMLInputElement>) {
+    setInfnt(evnt.target.checked);
+  }
```
VitaliyBondarenko1982, line 46: consider to make all inpurs fields as object where each field will be equal input name field
and after it will be possible to make one onChange handler for all fields instead five

```
  itemWidth: '',
  frameSize: '',
  ....
})

function changeAnimation(evnt: React.ChangeEvent<HTMLInputElement>) {
   const { name, value } = evnt.target;

    setInputs(prevINputs => ({ ...prevInputs, [name]: +value }));
  }
```

VitaliyBondarenko1982 (changes requested):
Looks good
Consider some improvements.

Will be better to limit some min and max value in inputs for avoid such behavior
Screenshot from 2023-07-11 16-00-08
Screenshot from 2023-07-11 15-59-48

etojeDenys (approved):
GJ

PR #441, opened 2023-07-11
React__carousel

Soi4An:
DEMO - https://Soi4An.github.io/react_carousel/

Problems of tests:
    1) should have picture size 130 px by default
    2) should change scroll text on 1 step
    3) should change width of image
    4) should change frame size

...but these propertes are correct working

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,5 @@
-// style go here
+.App {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,34 +3,47 @@ import './App.scss';
 import Carousel from './components/Carousel';
 
 interface State {
-  images: string[];
+  images: string[],
+  step: number,
+  frameSize: number,
+  itemWidth: number,
+  animationDuration: number,
+  infinite: boolean,
 }
 
 class App extends React.Component<{}, State> {
   state = {
-    images: [
-      './img/1.png',
-      './img/2.png',
-      './img/3.png',
-      './img/4.png',
-      './img/5.png',
-      './img/6.png',
-      './img/7.png',
-      './img/8.png',
-      './img/9.png',
-      './img/10.png',
-    ],
+    images: ['1', '2', '3', '4', '5', '6', '7', '8', '9', '10'],
+    step: 3,
+    frameSize: 3,
+    itemWidth: 130, // px
+    animationDuration: 1000, // 1000
+    infinite: false,
   };
 
   render() {
-    const { images } = this.state;
+    const {
+      images,
+      step,
+      frameSize,
+      itemWidth,
+      animationDuration,
+      infinite,
+    } = this.state;
 
     return (
       <div className="App">
         {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1 data-cy="title">Carousel with {images.length} images</h1>
 
-        <Carousel />
+        <Carousel
+          imagesNumbers={images}
+          step={step}
+          frameSize={frameSize}
+          itemWidth={itemWidth}
+          animationDuration={animationDuration}
+          infinite={infinite}
+        />
       </div>
     );
   }
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,109 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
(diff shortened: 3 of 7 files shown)


mykhalenych (changes requested):
Screenshot 2023-07-11 at 17 42 59
missed requirements, if you have questions, add them to chat

volodymyr-soltys97 (changes requested):
The carousel now spins by itself, you need to implement it so that it happens only when you click
https://github.com/mate-academy/react_carousel/assets/64750958/7516e46c-b731-444c-82d7-959d0baad4e6

What does mean `fasle`?
image

volodymyr-soltys97 (changes requested):
Not fixed from the previous review
image

Soi4An:
> The carousel now spins by itself, you need to implement it so that it happens only when you click https://github.com/mate-academy/react_carousel/assets/64750958/7516e46c-b731-444c-82d7-959d0baad4e6
> 
> What does mean `fasle`? image

it is mean that i have added describe (comments) in my project and i wanted to you read it

volodymyr-soltys97 (changes requested):
It is necessary to return this code and replace the word fasle with false
image

src/App.scss
```diff
@@ -1 +1,17 @@
-// style go here
```
anastasiiavorobiova, line 1: Line comments should not be used in CSS. Use block comments instead.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,96 @@
...
+          onClick={() => prevImages()}
+        >
+          Prev
+        </button>
+
+        <button
+          type="button"
+          data-cy="next"
+          disabled={infinite ? false : moveRight === lastPosition}
+          onClick={() => nextImages()}
```
anastasiiavorobiova, line 87: The 'Next' button is missing the 'title' attribute. It should be added for better accessibility.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,96 @@
...
+        </ul>
+      </div>
+
+      <div className="button__container">
+        <button
+          type="button"
+          disabled={infinite ? false : moveRight === 0}
+          onClick={() => prevImages()}
+        >
+          Prev
```
anastasiiavorobiova, line 80: The 'Prev' button is missing the 'title' attribute. It should be added for better accessibility.

src/App.tsx
```diff
@@ -3,34 +3,131 @@ import './App.scss';
...
+            step:
+            <input
+              className="values__input"
+              type="number"
+              name="step"
+              min="1"
+              max="5"
+              step="1"
+              value={step}
+              onChange={(event) => {
```
anastasiiavorobiova, line 54: Direct state mutation is not recommended in React. It would be advisable to create a single method that updates the state based on the input's name and value. This method can be reused for all inputs to avoid redundancy.

src/App.tsx
```diff
@@ -3,34 +3,131 @@ import './App.scss';
...
+              name="frameSize"
+              min="130"
+              max="300"
+              step="10"
+              value={itemWidth}
+              onChange={(event) => {
+                this.setState({ itemWidth: +event.target.value });
+              }}
+            />
+          </label>
```
anastasiiavorobiova, line 90: For accessibility, it is good to provide a default value for the 'select' element. Currently, no option is selected by default.

src/App.tsx
```diff
@@ -3,34 +3,131 @@ import './App.scss';
...
-      './img/3.png',
-      './img/4.png',
-      './img/5.png',
-      './img/6.png',
-      './img/7.png',
-      './img/8.png',
-      './img/9.png',
-      './img/10.png',
-    ],
+    images: ['1', '2', '3', '4', '5', '6', '7', '8', '9', '10'],
```
anastasiiavorobiova, line 16: The 'images' state array should contain paths to images, not just numbers as it is now. Please, correct this.

src/App.tsx
```diff
@@ -3,34 +3,131 @@ import './App.scss';
...
+              }}
+            >
+              <option value="false">false</option>
+              <option value="true">true</option>
+            </select>
+          </label>
+        </form>
+
+        <Carousel
+          imagesNumbers={images}
```
anastasiiavorobiova, line 123: Props of the 'Carousel' component don't match the requirements. Instead of 'imagesNumbers', it should be 'images' prop with an array of image URLs.

src/App.tsx
```diff
@@ -3,34 +3,131 @@ import './App.scss';
...
+              }}
+            />
+          </label>
+
+          <label className="values__label">
+            itemWidth (px):
+            <input
+              className="values__input"
+              type="range"
+              name="frameSize"
```
anastasiiavorobiova, line 81: The 'name' attribute of the input field should be 'itemWidth', not 'frameSize'. Make sure all input fields have correct 'name' attribute corresponding to their purpose.

anastasiiavorobiova (changes requested):
Great work! Added comments to fix

PR #442, opened 2023-07-11
Develop

VolodymyrPst:
https://volodymyrpst.github.io/react_carousel/

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,5 @@
-// style go here
+.App {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import './App.scss';
 import Carousel from './components/Carousel';
+import { Value } from './types.tsx/Values';
 
 interface State {
   images: string[];
@@ -20,17 +21,45 @@ class App extends React.Component<{}, State> {
       './img/9.png',
       './img/10.png',
     ],
+    itemWidth: '130',
+    frameSize: '3',
+    step: '2',
+    animationDuration: '1000',
+    infinite: false,
+  };
+
+  stateHandler = (value: Value) => {
+    this.setState(state => ({
+      images: state.images,
+      ...value,
+    }));
   };
 
   render() {
-    const { images } = this.state;
+    const {
+      images,
+      itemWidth,
+      frameSize,
+      step,
+      animationDuration,
+      infinite,
+    } = this.state;
 
     return (
       <div className="App">
-        {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1>
+          {`Carousel with ${images.length} images`}
+        </h1>
 
-        <Carousel />
+        <Carousel
+          images={images}
+          step={step}
+          frameSize={frameSize}
+          itemWidth={itemWidth}
+          animationDuration={animationDuration}
+          infinite={infinite}
+          chengeState={this.stateHandler}
+        />
       </div>
     );
   }
--- a/src/components/Carousel.scss
+++ b/src/components/Carousel.scss
@@ -1,1 +1,33 @@
-// style go here
+.Carousel {
+  width: 1300px;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+
+  &__wrapper {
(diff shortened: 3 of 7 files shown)


polosanya (changes requested):
1. Cyclic doesn't want
image
2. It would be great to immediately display the new item width and frame size.
image
Also, it is possible to set the negative values

vlad-demchuk (changes requested):
1. you didn't implement this point
Add a disabled class and style a button with this class accordingly if clicking the button would have no effect (i.e., there are no more images to display).

2. if you have Infinite option enabled, it should return to the first image after the last one
3. also, set some maximum values to avoid bugs
image

TarasHoliuk:
1. It shouldn't scroll to this position - the third (last) element is empty:
image

2. No need to set `disabled` styles for buttons if `infinity` is enabled:
image

3. Only previously set value working:
/home/vestr/Videos/Screencasts/Screencast from 16-07-23 11:21:10.webm

src/components/Carousel.tsx
```diff
@@ -1,18 +1,196 @@
...
+  const clickHandler = () => {
+    const value = {
+      step: steps,
+      frameSize: frame,
+      itemWidth: itemWdt,
+      animationDuration: duration,
+      infinite: cyclic,
+    };
+
+    chengeState(value);
```
TarasHoliuk, line 43:
```suggestion
    changeState(value);
```
typo

src/App.tsx
```diff
@@ -20,17 +21,45 @@ class App extends React.Component<{}, State> {
       './img/9.png',
       './img/10.png',
     ],
+    itemWidth: '130',
+    frameSize: '3',
+    step: '2',
+    animationDuration: '1000',
+    infinite: false,
```
TarasHoliuk, line 28: no need to save it in the state here and in the `Carousel` component
Keep it here and render inputs here instead of the `Carousel` component

src/App.tsx
```diff
@@ -20,17 +21,45 @@ class App extends React.Component<{}, State> {
...
+  };
+
+  stateHandler = (value: Value) => {
+    this.setState(state => ({
+      images: state.images,
```
TarasHoliuk, line 33: it was an example
actually you can move out images out from the state since it's static value

TarasHoliuk (changes requested):

src/App.scss
```diff
@@ -1 +1,5 @@
-// style go here
```
volodymyr-soltys97, line 1: The comment 'style go here' is not necessary. It is a good practice to remove unnecessary comments to keep the code clean.

src/components/Carousel.scss
```diff
@@ -1 +1,40 @@
...
+    flex-direction: column;
+    margin: 0 10px;
+  }
+
+  &__btn {
+    background: rgba(102, 102, 247, 0.55);
+  }
+
+  &__end {
+    background: rgb(106, 106, 145);
```
volodymyr-soltys97, line 38: Colors should be stored in variables for consistency and ease of maintenance. Consider creating a variable for this color value.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,180 @@
...
+          className="Carousel__list"
+          style={{
+            transform: `translateX(${translate}px)`,
+            transition: `transform ${(animationDuration)}ms`,
+          }}
+        >
+          {
+            images.map((img, index) => {
+              return (
+                <li>
```
volodymyr-soltys97, line 83: The 'key' prop should be defined on the 'li' element itself, not on the 'img' child.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,180 @@
...
+  images,
+  step,
+  frameSize,
+  itemWidth,
+  animationDuration,
+  infinite,
+  changeState,
+}) => {
+  const interval = useRef<NodeJS.Timeout | null>(null);
+  const [translate, setTranslate] = useState(0);
```
volodymyr-soltys97, line 26: Consider renaming 'translate' state variable to something more meaningful like 'carouselPosition'.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

volodymyr-soltys97 (approved):
Well done 🔥

PR #459, opened 2023-07-26
Solution

kateryna-stemblivska:
[DEMO LINK](https://kateryna-stemblivska.github.io/react_carousel/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,38 @@
-// style go here
+.App {
+  width: max-content;
+  margin: 50px auto;
+  padding: 30px;
+  display: flex;
+  justify-content: center;
+  flex-direction: column;
+  align-items: center;
+  gap: 50px;
+  border: 5px solid rgba(182, 32, 6, 0.97);
+  border-radius: 25px;
+
+  &__title {
+    margin: 0;
+    font-size: 40px;
+    color: rgba(182, 32, 6, 0.97);
+  }
+}
+
+.settings {
+  width: 600px;
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  gap: 20px;
+
+  &__label {
+    font-size: 18px;
+    font-style: italic;
+
+  }
+
+  &__input {
+    width: 60px;
+    margin-left: 5px;
+
+  }
+}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import Carousel from './components/Carousel';
 
@@ -6,8 +6,8 @@ interface State {
   images: string[];
 }
 
-class App extends React.Component<{}, State> {
-  state = {
+const App: React.FC = () => {
+  const state: State = {
     images: [
       './img/1.png',
       './img/2.png',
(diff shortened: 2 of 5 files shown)


src/components/Carousel.scss
```diff
@@ -1 +1,53 @@
...
+    align-items: center;
+    column-gap: 50px;
+  }
+
+  &__button {
+    width: 75px;
+    padding: 2px 0;
+    border: 2px solid rgba(182, 32, 6, 0.97);
+    border-radius: 25px;
+    background-color: yellow;
```
maxim2310, line 30: Same as above, consider defining the color as a variable.

src/App.scss
```diff
@@ -1 +1,38 @@
...
+.App {
+  width: max-content;
+  margin: 50px auto;
+  padding: 30px;
+  display: flex;
+  justify-content: center;
+  flex-direction: column;
+  align-items: center;
+  gap: 50px;
+  border: 5px solid rgba(182, 32, 6, 0.97);
```
maxim2310, line 10: You should avoid using direct color values. Consider using a CSS variable for the color rgba(182, 32, 6, 0.97) and refer to this variable whenever this color is needed. It will make your code easier to maintain and provide a consistent color scheme.

src/App.tsx
```diff
@@ -1,13 +1,13 @@
...
 import Carousel from './components/Carousel';
 
 interface State {
   images: string[];
 }
 
-class App extends React.Component<{}, State> {
-  state = {
+const App: React.FC = () => {
+  const state: State = {
```
maxim2310, line 10: The 'state' variable is not a react state but is named as such which can be confusing. Consider renaming it to something more descriptive

src/App.tsx
```diff
@@ -22,18 +22,118 @@ class App extends React.Component<{}, State> {
...
+        <label
+          htmlFor="itemId"
+          className="settings__label"
+        >
+          Set item width:
+          <input
+            className="settings__input"
+            type="number"
+            id="itemId"
+            defaultValue={itemWidth}
```
maxim2310, line 48: Using 'defaultValue' instead of 'value' in controlled components can lead to unexpected behavior. Use 'value' prop instead for react controlled components.

src/App.tsx
```diff
@@ -22,18 +22,118 @@ class App extends React.Component<{}, State> {
...
+        <label
+          htmlFor="stepId"
+          className="settings__label"
+        >
+          Set step:
+          <input
+            className="settings__input"
+            type="number"
+            id="stepId"
+            defaultValue={step}
```
maxim2310, line 84: Using 'defaultValue' instead of 'value' in controlled components can lead to unexpected behavior. Use 'value' prop instead for react controlled components.

src/App.tsx
```diff
@@ -22,18 +22,118 @@ class App extends React.Component<{}, State> {
...
+        <label
+          htmlFor="frameId"
+          className="settings__label"
+        >
+          Set frame size:
+          <input
+            className="settings__input"
+            type="number"
+            id="frameId"
+            defaultValue={frameSize}
```
maxim2310, line 66: Using 'defaultValue' instead of 'value' in controlled components can lead to unexpected behavior. Use 'value' prop instead for react controlled components.

src/App.tsx
```diff
@@ -22,18 +22,118 @@ class App extends React.Component<{}, State> {
...
   };
 
-  render() {
-    const { images } = this.state;
+  const { images } = state;
+  const [step, setStep] = useState(3);
+  const [frameSize, setFrameSize] = useState(3);
+  const [itemWidth, setItemWidth] = useState(130);
+  const [animationDuration, setAnimationDuration] = useState(1000);
+  const [toggleChecked, setToggleChecked] = useState(false);
```
maxim2310, line 30: The variable 'toggleChecked' is not clearly named for its purpose. Consider renaming it to 'isInfinite' or 'infiniteAnimation' to better reflect its function.

src/App.tsx
```diff
@@ -22,18 +22,118 @@ class App extends React.Component<{}, State> {
...
+        <label
+          htmlFor="animationId"
+          className="settings__label"
+        >
+          Set animation duration:
+          <input
+            className="settings__input"
+            type="number"
+            id="animationId"
+            defaultValue={animationDuration}
```
maxim2310, line 102: Using 'defaultValue' instead of 'value' in controlled components can lead to unexpected behavior. Use 'value' prop instead for react controlled components.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,87 @@
...
+  const widthOfListBox = itemWidth * frameSize;
+  const hiddenImagesLength = (images.length - frameSize) * itemWidth;
+
+  const getPrevImage = () => {
+    setCurrentShift((shift) => Math.min(shift + itemWidth * step, 0));
+  };
+
+  const getNextImage = () => {
+    if (currentShift === -hiddenImagesLength && infinite) {
+      setCurrentShift(0);
```
maxim2310, line 32: When the carousel is in the 'infinite' mode and you are at the end of the images, you are setting the 'currentShift' to 0, which would abruptly shift the carousel to the start instead of scrolling smoothly. A better approach would be to duplicate the images such that it appears to be infinite scrolling.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,87 @@
...
+
+  return (
+    <div className="Carousel">
+      <ul
+        className="Carousel__list"
+        style={{ width: `${widthOfListBox}px` }}
+      >
+        {images.map((img, ind) => (
+          <li
+            key={img}
```
maxim2310, line 49: Using image URL as a key may cause problems if you have duplicate images. It's better to use something unique like the index or a combination of index and url.

maxim2310 (changes requested):
Nice work,
need to make a small fixes
and find for you one bug)
https://www.loom.com/share/4e202c24d7294a199a95f677dc8c9cdd

src/App.scss
```diff
@@ -1 +1,40 @@
...
```
witflash, line 12: Here you can use your `main-color` variable too

witflash (approved):
Well done 👍🏻

PR #465, opened 2023-07-27
Solution

artyomwhite:
[DEMO LINK](https://artyomwhite.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,12 +2,17 @@ import React from 'react';
 import './App.scss';
 import Carousel from './components/Carousel';
 
-interface State {
+type State = {
   images: string[];
-}
+  step: number,
+  frameSize: number,
+  itemWidth: number,
+  animationDuration: number,
+  infinite: boolean,
+};
 
 class App extends React.Component<{}, State> {
-  state = {
+  state: Readonly<State> = {
     images: [
       './img/1.png',
       './img/2.png',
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -20,17 +25,108 @@ class App extends React.Component<{}, State> {
...
+              onChange={this.handleChange}
+            />
+          </label>
+          <label htmlFor="frameSize" className="App__label">
+            Frame size:
+            <input
+              id="frameSize"
+              type="number"
+              className="App__input"
+              defaultValue={3}
```
Viktor-Kost, line 88: The defaultValue for the frameSize input field is hardcoded as 3. It would be better to use the frameSize from the state as the defaultValue. Check other values that you can take from the state

src/App.tsx
```diff
@@ -20,17 +25,108 @@ class App extends React.Component<{}, State> {
...
+              onChange={this.handleChange}
+            />
+          </label>
+          <label htmlFor="infinite" className="App__label">
+            Infinite:
+            <input
+              id="infinite"
+              type="checkbox"
+              className="App__input"
+              onChange={(event) => this.setState({
```
Viktor-Kost, line 124: The onchange event for the infinite input field is setting the state directly. It would be better to use the handleChange method for this to keep the state updates consistent.

Viktor-Kost (changes requested):
GJ!
Some minor fixes needed

mykhalenych (approved):

PR #471, opened 2023-08-02
task solution

Bonum-taurus:
https://Bonum-taurus.github.io/react_carousel/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,6 +4,11 @@ import Carousel from './components/Carousel';
 
 interface State {
   images: string[];
+  frameSize: number;
+  itemWidth: number;
+  step: number,
+  animationDuration: number,
+  infinite: boolean,
 }
 
 class App extends React.Component<{}, State> {
@@ -20,17 +25,78 @@ class App extends React.Component<{}, State> {
       './img/9.png',
       './img/10.png',
     ],
+    frameSize: 3,
+    itemWidth: 130,
+    step: 3,
+    animationDuration: 1000,
+    infinite: false,
+  };
+
+  changeItemWidth = (width: number) => {
+    this.setState({ itemWidth: width });
+  };
+
+  changeFrameSize = (size: number) => {
+    this.setState({ frameSize: size });
+  };
+
+  changeStepLong = (long: number) => {
+    this.setState({ step: long });
+  };
+
+  changeDuration = (time: number) => {
+    this.setState({ animationDuration: time });
   };
 
   render() {
-    const { images } = this.state;
+    const {
+      images,
+      frameSize,
+      itemWidth,
+      step,
+      animationDuration,
+      infinite,
+    } = this.state;
 
     return (
       <div className="App">
         {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1 data-cy="title">Carousel with {images.length} images</h1>
 
-        <Carousel />
+        <input
+          type="text"
+          className="field"
+          placeholder="item width"
+          onChange={(event) => this.changeItemWidth(+event.currentTarget.value)}
+        />
+        <input
+          type="text"
+          className="field"
+          placeholder="frame size"
+          onChange={(event) => this.changeFrameSize(+event.currentTarget.value)}
+        />
+        <input
+          type="text"
+          className="field"
+          placeholder="step"
+          onChange={(event) => this.changeStepLong(+event.currentTarget.value)}
+        />
+
+        <input
+          type="text"
+          className="field"
+          placeholder="animation duration"
+          onChange={(event) => this.changeDuration(+event.currentTarget.value)}
+        />
+
+        <Carousel
+          images={images}
+          frameSize={frameSize}
+          itemWidth={itemWidth}
+          step={step}
+          animationDuration={animationDuration}
+          infinite={infinite}
+        />
       </div>
     );
   }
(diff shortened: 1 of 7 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your carousel.
1. You need to show the default value in these fields
image

2. It will be better if you place the carousel in the center of the page
image

volodymyr-soltys97 (changes requested):
Also add a description to these fields at the top so that the user understands what each field is responsible for
image

src/components/Carousel.scss
```diff
@@ -1 +1,11 @@
-// style go here
+.Carousel {
+  margin: 0 auto;
+  text-align: center;
+  
+  overflow: hidden;
+}
+
+[type="button"] {
```
volodymyr-soltys97, line 8: The [type='button'] selector is too general. This will affect all button elements, not just those in the Carousel component. Consider making this more specific, for example, by nesting it under the .Carousel class.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,98 @@
...
+
+            if (currentShift > -(itemWidth * step)) {
+              return 0;
+            }
+
+            return currentShift + itemWidth * step;
+          });
+        }}
+        type="button"
+        disabled={shift >= 0 && !infinite && true}
```
volodymyr-soltys97, line 68: There is an unnecessary boolean operation in the disabled prop. When shift >= 0 && !infinite evaluates to true, there is no need to add another && true. This can be simplified.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,98 @@
...
+              + itemWidth * step) {
+              return -(widthCarousel - widthPortion);
+            }
+
+            return currentShift - itemWidth * step;
+          });
+        }}
+        data-cy="Next"
+        type="button"
+        disabled={shift <= -(widthCarousel - widthPortion) && !infinite && true}
```
volodymyr-soltys97, line 90: There is an unnecessary boolean operation in the disabled prop. When shift <= -(widthCarousel - widthPortion) && !infinite evaluates to true, there is no need to add another && true. This can be simplified.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,98 @@
...
+};
+
+const Carousel: React.FC<Props> = ({
+  images,
+  frameSize,
+  itemWidth,
+  step,
+  animationDuration,
+  infinite,
+}) => {
```
volodymyr-soltys97, line 20: The default values of frameSize, itemWidth, step, animationDuration are missing. You should provide default values to these props using defaultProps or setting default values in functional component's arguments.

src/App.tsx
```diff
@@ -20,17 +25,106 @@ class App extends React.Component<{}, State> {
...
+  changeItemWidth = (width: number) => {
+    this.setState({ itemWidth: width });
+  };
+
+  changeFrameSize = (size: number) => {
+    this.setState({ frameSize: size });
+  };
+
+  changeStepLong = (long: number) => {
+    this.setState({ step: long });
```
volodymyr-soltys97, line 44: You should validate the input before setting the state to prevent potential bugs or crashes.

src/App.tsx
```diff
@@ -20,17 +25,106 @@ class App extends React.Component<{}, State> {
...
+    animationDuration: 1000,
+    infinite: false,
+  };
+
+  changeItemWidth = (width: number) => {
+    this.setState({ itemWidth: width });
+  };
+
+  changeFrameSize = (size: number) => {
+    this.setState({ frameSize: size });
```
volodymyr-soltys97, line 40: You should validate the input before setting the state to prevent potential bugs or crashes.

src/App.tsx
```diff
@@ -20,17 +25,106 @@ class App extends React.Component<{}, State> {
...
     ],
+    frameSize: 3,
+    itemWidth: 130,
+    step: 3,
+    animationDuration: 1000,
+    infinite: false,
+  };
+
+  changeItemWidth = (width: number) => {
+    this.setState({ itemWidth: width });
```
volodymyr-soltys97, line 36: You should validate the input before setting the state to prevent potential bugs or crashes.

src/App.tsx
```diff
@@ -20,17 +25,106 @@ class App extends React.Component<{}, State> {
...
+  changeFrameSize = (size: number) => {
+    this.setState({ frameSize: size });
+  };
+
+  changeStepLong = (long: number) => {
+    this.setState({ step: long });
+  };
+
+  changeDuration = (time: number) => {
+    this.setState({ animationDuration: time });
```
volodymyr-soltys97, line 48: You should validate the input before setting the state to prevent potential bugs or crashes.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,98 @@
...
+              return 0;
+            }
+
+            if (currentShift <= -(widthCarousel - widthPortion)
+              + itemWidth * step) {
+              return -(widthCarousel - widthPortion);
+            }
+
+            return currentShift - itemWidth * step;
+          });
```
volodymyr-soltys97, line 86: It would be a good practice to put such logic into a separate function and then use it there, instead of writing such extensive logic in `jsx`

volodymyr-soltys97 (changes requested):
Let's improve your code

src/components/Carousel.tsx
Bonum-taurus: Я прописав defaultValue для інпутів.

volodymyr-soltys97 (approved):
Finally, well done 🔥

PR #474, opened 2023-08-04
solution

HYPER2307:
[DEMO LINK](https://hyper2307.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,7 +3,11 @@ import './App.scss';
 import Carousel from './components/Carousel';
 
 interface State {
-  images: string[];
+  images: string[],
+  step: number,
+  frameSize: number,
+  animationDuration: number,
+  infinity: boolean,
 }
 
 class App extends React.Component<{}, State> {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -20,17 +24,116 @@ class App extends React.Component<{}, State> {
       './img/9.png',
       './img/10.png',
     ],
```
VitaliyBondarenko1982, line 26: looks like no need to keep images in state. it is can be constant outside of component

src/components/Carousel.tsx
```diff
@@ -1,18 +1,101 @@
...
+            />
+          </li>
+        ))}
+      </ul>
+
+      <button
+        type="button"
+        onClick={() => (
+          setOffset(handlePrevClick)
+        )}
```
VitaliyBondarenko1982, line 84: consider to make one handler factory which will except callback and return arrow function.
```
const handleOffset = (callback: React.Dispatch<React.SetStateAction<number>>) => () => {
  setOffset(callback)
}
```
```suggestion
        onClick={handleOffset(handlePrevClick)}
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,101 @@
...
+      >
+        Prev
+      </button>
+      <button
+        type="button"
+        data-cy="next"
+        onClick={() => (
+          setOffset(handleNextClick)
+        )}
```
VitaliyBondarenko1982, line 93:
```suggestion
        onClick={handleOffset(handleNextClick)}
```

VitaliyBondarenko1982 (changes requested):
Consider to limit min values for some fields for avoid such view.
and prevent to input 0.
Screenshot from 2023-08-04 12-53-29
Screenshot from 2023-08-04 12-52-48

etojeDenys (approved):
good job

PR #475, opened 2023-08-06
Solution

katerinalex:
[DEMO LINK](https://katerinalex.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,90 @@
 import React from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
-      <li><img src="./img/1.png" alt="3" /></li>
-      <li><img src="./img/1.png" alt="4" /></li>
-    </ul>
+// const Carousel: React.FC = () => (
+//   <div className="Carousel">
+//     <ul className="Carousel__list">
+//       <li><img src="./img/1.png" alt="1" /></li>
+//       <li><img src="./img/1.png" alt="2" /></li>
+//       <li><img src="./img/1.png" alt="3" /></li>
+//       <li><img src="./img/1.png" alt="4" /></li>
+//     </ul>
 
-    <button type="button">Prev</button>
-    <button type="button">Next</button>
-  </div>
-);
+//     <button type="button">Prev</button>
+//     <button type="button">Next</button>
+//   </div>
+// );
 
-export default Carousel;
+type Props = {
+  images: string[],
+  step: number,
+  frameSize: number,
+  itemWidth: number,
+  animationDuration: number,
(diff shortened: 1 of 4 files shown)


maxim2310 (changes requested):
Most likely you didnt deploy your last version, do it, and write me again 
image

src/components/Carousel.tsx
```diff
@@ -1,18 +1,97 @@
...
+      } else {
+        setFirstLiStyle(prevStyle => ({
+          ...prevStyle,
+          marginLeft: '0',
+        }));
+        setCurrent(0);
+      }
+    }
+
+    // console.log('handlePrev:', firstLiStyle.marginLeft);
```
Anastasiia-Svintsova, line 52: Remove redundant commented code

Anastasiia-Svintsova (changes requested):

src/components/Carousel.tsx
```diff
@@ -1,18 +1,95 @@
...
+            ? (
+              <li key={image} style={firstLiStyle}>
+                <img className="image" style={imageStyle} src={image} alt={`${images.indexOf(image)}`} />
+              </li>
+            )
+            : (
+              <li key={image}>
+                <img className="image" style={imageStyle} src={image} alt={`${images.indexOf(image)}`} />
+              </li>
+            )
```
Anastasiia-Svintsova, line 86: You don't need to repeat the same code twice. You can move a condition inside the style attribute

src/components/Carousel.tsx
```diff
@@ -1,18 +1,95 @@
...
+          marginLeft: `${(parseInt(prevStyle.marginLeft, 10)) + (step * itemWidth)}px`,
+        }));
+        setCurrent(old => old - frameSize);
+      } else {
+        setFirstLiStyle(prevStyle => ({
+          ...prevStyle,
+          marginLeft: '0',
+        }));
+        setCurrent(0);
+      }
```
Anastasiia-Svintsova, line 49: No need to write it twice. You can refactor it by using the ternary operator to simplify the code

Anastasiia-Svintsova (changes requested):
- Disable buttons if the user can't scroll anymore 

https://github.com/mate-academy/react_carousel/assets/99403198/b4c65519-0716-484b-8919-fa62df2493f4

- Would be great to add some styles for buttons to make the UI look better. It doesn't look professional enough without any styles

Anastasiia-Svintsova (changes requested):
- There are no style changes when buttons become disabled. How the user should understand it? And the cursor should be `pointer` to show that the user can interact with the button.

- Steps work wrong. Watch the video below 

https://github.com/mate-academy/react_carousel/assets/99403198/a6a64bb7-3d6a-470a-95d8-ecb0844cc1eb

sTorba24 (changes requested):
Almost good! lets make it perfect

1. When i change item size after some scrolling i see it(use pointer to check which item should be shown first) (like firstImageIndex)
image

2. Would be nice to implement `inifinite` option

witflash (approved):
Finally great work!
Well done 👍

PR #478, opened 2023-08-08
Solution

AnnaViolin23:
[DEMO LINK](https://AnnaViolin23.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,10 +19,20 @@ const images = [
 
 const imageCount = images.length;
 
-const App: React.FC = () => {
-  const defaultItemWidth = window.innerWidth < 390 ? 50 : 130;
+const calculateDefaultItemWidth = (): number => {
+  let defaultItemWidth = 130;
 
-  const [itemWidth, setItemWidth] = useState(defaultItemWidth);
+  if (window.innerWidth < 500) {
+    defaultItemWidth = 60;
+  } else if (window.innerWidth < 690) {
+    defaultItemWidth = 100;
+  }
+
+  return defaultItemWidth;
+};
+
+const App: React.FC = () => {
+  const [itemWidth, setItemWidth] = useState(calculateDefaultItemWidth);
   const [frameSize, setFrameSize] = useState(3);
   const [step, setStep] = useState(3);
   const [animationDuration, setAnimationDuration] = useState(1000);


src/components/Settings/Settings.scss
```diff
@@ -0,0 +1,58 @@
...
+  align-items: center;
+  gap: 15px;
+  max-width: 700px;
+  margin: 0 auto 40px;
+
+  &__wrapper {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    // width: 50vh;
```
Viktor-Kost, line 13: remove comments

src/App.tsx
```diff
@@ -1,39 +1,104 @@
...
   };
 
-  render() {
-    const { images } = this.state;
+  useEffect(() => {
+    setIsPrevButtontnActive(position > FIRST_POSITION);
+    setIsNextButtonActive(position < LAST_POSITION);
+  }, [position]);
+
+  useEffect(() => { }, [
```
Viktor-Kost, line 58: There is an empty useEffect. If it is not used, consider removing it to clean up your code.

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,97 @@
...
+  images: string[];
+  position: number,
+  animationDuration: number,
+  frameSize: number,
+  actualImageWidth: number,
+  isPrevButtonActive: boolean,
+  isNextButtonActive: boolean,
+  hadlePrevButtonClick: () => void
+  hadleNextButtonClick: () => void
+  getImageId: (url: string) => string;
```
Viktor-Kost, line 16: The getImageId function prop seems unnecessary. You can use the image url itself as the key in your map function.

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,97 @@
...
+                className="Carousel__item"
+                style={{
+                  flexBasis: `${actualImageWidth}px`,
+                }}
+              >
+                <img
+                  src={image}
+                  alt={`${getImageId(image)}`}
+                  style={{
+                    flexBasis: `${actualImageWidth}px`,
```
Viktor-Kost, line 60: Setting the flexBasis for the img seems unnecessary as it's already set for the parent li element.

src/App.tsx
```diff
@@ -1,39 +1,104 @@
...
+};
+
+const App: React.FC = () => {
+  const [itemWidth, setItemWidth] = useState(130);
+  const [frameSize, setFrameSize] = useState(3);
+  const [step, setStep] = useState(3);
+  const [animationDuration, setAnimationDuration] = useState(1000);
+  const [position, setPosition] = useState(0);
+  const [isPrevButtontnActive, setIsPrevButtontnActive] = useState(false);
+  const [isNextButtonActive, setIsNextButtonActive] = useState(true);
```
Viktor-Kost, line 33: As a recommendation for the future.
Since you have a lot of different options it'd be better to use it as one state
```
const [options, setOptions] = useState({
  itemWidth: 30,
  frameSize: 3,
  step: 3,
  animationDuration: 1000,
});
```

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,97 @@
...
+          <ul
+            className="Carousel__list"
+            style={{
+              transform: `translateX(${translateCoords}px)`,
+              transition: `transform ${animationDuration}ms linear`,
+            }}
+          >
+            {images.map((image: string) => (
+              <li
+                key={getImageId(image)}
```
Viktor-Kost, line 50: getImageId can return an empty string, so in theory you can get two items with the same key that isn't good

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,97 @@
...
+            })}
+            onClick={hadlePrevButtonClick}
+          >
+            Back
+          </button>
+          <button
+            type="button"
+            disabled={!isNextButtonActive}
+            className={cn('Carousel__button__next', {
+              'is-disabled': !isNextButtonActive,
```
Viktor-Kost, line 84: change styles a little for disabled buttons (as an example you can use some opacity)

Viktor-Kost (changes requested):
Nice!)
Some moments can be improved

volodymyr-soltys97 (approved):
Nice work 🔥

PR #479, opened 2023-08-10
initial basic carousel made

domkab:
https://domkab.github.io/react_carousel

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,13 @@
-// style go here
+.App {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  text-align: center;
+  height: 50vh;
+  justify-content: center;
+
+  &__title {
+    text-align: center;
+    // margin: 0;
+  }
+}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -28,9 +28,24 @@ class App extends React.Component<{}, State> {
     return (
       <div className="App">
         {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1
+          className="App__title"
+        >
+          Carousel with
+          {' '}
+          {images.length}
+          {' '}
+          images
 
-        <Carousel />
+        </h1>
+
+        <Carousel
+          images={images}
+          frameSize={2}
+          step={3}
+          itemWidth={200}
+          animationDuration={1000}
+        />
       </div>
     );
   }
--- a/src/components/Carousel.scss
+++ b/src/components/Carousel.scss
@@ -1,1 +1,89 @@
-// style go here
+:root {
+  --image-size: 130px;
+  --animation-duration: 1s;
+  --transform-offset: 0;
+}
+
+.disabled {
+  cursor: not-allowed;
+  opacity: 0.5;
+  pointer-events: none;
+}
+
+img {
+  width: var(--image-size);
+  height: var(--image-size);
+}
+
+.transformed {
+  transform: translateX(var(--transform-offset));
+}
+
+.prev-button,
+.next-button {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  background-color: #eee;
+  padding: 10px 15px;
+  border: none;
+  cursor: pointer;
+  position: absolute;
+}
+
+.prev-button {
+  left: 0;
+  &::before {
+    content: "";
+    display: block;
+    margin-right: 5px;
+    width: 8px;
+    height: 8px;
+    border: solid 2px;
+    border-right: none;
+    border-bottom: none;
+    transform: rotate(-45deg);
+  }
+}
+
+.next-button {
+  right: 0;
+  &::after {
+    content: "";
+    display: block;
+    margin-left: 5px;
+    width: 8px;
+    height: 8px;
+    border: solid 2px;
+    border-left: none;
+    border-top: none;
+    transform: rotate(-45deg);
+  }
+}
+
+.Carousel {
+  position: relative;
+  display: flex;
+  align-items: center;
+  width: calc(var(--frame-size) * var(--image-size));
+  padding: 0 30px;
+  overflow: hidden;
+
+  &__list {
+    display: flex;
+    list-style: none;
+    transform: translateX(var(--transform-offset));
+    transition: transform var(--animation-duration) ease-out;
+  }
+
+  & li:not(:last-child) img {
+    margin-right: 10px;
+  }
+}
+
+.container {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100%;
+}
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,77 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
-      <li><img src="./img/1.png" alt="3" /></li>
-      <li><img src="./img/1.png" alt="4" /></li>
-    </ul>
+type CarouselProps = {
+  images: string[];
+  step: number;
+  frameSize: number;
+  itemWidth: number;
+  animationDuration: number;
+  // infinite?: boolean;
+};
 
-    <button type="button">Prev</button>
-    <button type="button">Next</button>
-  </div>
-);
+const Carousel: React.FC<CarouselProps> = ({
+  images,
+  frameSize = 3,
+  itemWidth = 130,
+  step = 3,
+  animationDuration = 1000,
+}) => {
+  const [offset, setOffset] = useState(0);
+  const imagesLeft = images.length - offset;
+
+  const handlePrev = () => {
+    if (offset >= step) {
+      setOffset(prevOffset => prevOffset - step);
+    } else {
+      setOffset(0);
+    }
+  };
+
+  const handleNext = () => {
+    if (offset + step < images.length - frameSize + 1) {
+      setOffset(prevOffset => prevOffset + step);
+    }
+  };
+
+  useEffect(() => {
+    const scrollWidth = offset * itemWidth;
+
+    document.documentElement.style.setProperty('--transform-offset', `-${scrollWidth}px`);
+    document.documentElement.style.setProperty('--image-size', `${itemWidth}px`);
+    document.documentElement.style.setProperty('--frame-size', `${frameSize}`);
+    document.documentElement.style.setProperty('--animation-duration', `${animationDuration}ms`);
+  }, [offset, itemWidth, frameSize, animationDuration]);
+
+  return (
+    <div className="container">
+      <div className="Carousel">
+        <ul className="Carousel__list transformed">
+          {images.map((imgSrc) => (
+            <li key={imgSrc}>
+              <img src={imgSrc} alt="" />
+            </li>
+          ))}
+        </ul>
+
+        <button
+          type="button"
+          onClick={handlePrev}
+          className={`prev-button ${offset === 0 ? 'disabled' : ''}`}
+        >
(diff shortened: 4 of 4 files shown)


src/App.tsx
```diff
@@ -28,9 +28,24 @@ class App extends React.Component<{}, State> {
...
         {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1
+          className="App__title"
+        >
+          Carousel with
+          {' '}
+          {images.length}
+          {' '}
+          images
```
Anastasiia-Svintsova, line 38: Consider using [Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) for better readability

src/components/Carousel.scss
```diff
@@ -1 +1,89 @@
...
+  --transform-offset: 0;
+}
+
+.disabled {
+  cursor: not-allowed;
+  opacity: 0.5;
+  pointer-events: none;
+}
+
+img {
```
Anastasiia-Svintsova, line 13: Use classes for styles

Anastasiia-Svintsova (changes requested):
Keep going!

- Buttons shouldn't overlap images 
image

- The `Next` button should become disabled when the user couldn't scroll anymore

maxim2310 (changes requested):
Good work, but lets have some bugs, lets fix them: 
1. need to add input for quantity img in frame and set it without cutting
image
2.now your input have no effect on carousel

Anastasiia-Svintsova (changes requested):
- The images should start from the beginning. Don't cut off the last image 
image

- I can't see the last image but the next button is disabled already. Make sure that you calculate everything correctly 
image

- Add min values for inputs 
image

src/App.tsx
```diff
@@ -27,10 +27,21 @@ class App extends React.Component<{}, State> {
...
+          className="App__title"
+          data-cy="title"
+        >
+          {`Carousel with ${images.length} images`}
+        </h1>
 
-        <Carousel />
+        <Carousel
+          images={images}
+          frameSize={3}
```
lerastarynets, line 39: Hardcoded values for the Carousel's props. The task requires inputs for `itemWidth`, `frameSize`, `step`, and `animationDuration`. Consider having these values as part of the state and creating corresponding input fields to allow users to customize the carousel's behavior.

src/styles/blocks/button.scss
```diff
@@ -0,0 +1,73 @@
...
+    background-color: #2a3fb8ee;
+    color: white;
+  }
+
+  &:active {
+    transform: scale(1.2);
+  }
+}
+
+// input {
```
lerastarynets, line 67: Commented code detected. It's a good practice to remove code that is not in use. If you want to provide some information, you should write this in form of a comment, not in commented code.

src/styles/blocks/button.scss
```diff
@@ -0,0 +1,73 @@
...
+//   // height: 40px;
+// }
+
+input {
```
lerastarynets, line 71: The input styles are not nested within a parent selector. This could cause styles to leak and affect other elements in the application. Consider nesting these styles inside a specific parent selector to increase specificity and avoid potential CSS leaks.

src/styles/blocks/controls.scss
```diff
@@ -0,0 +1,21 @@
+.controls {
+  display: flex;
+  justify-content: space;
```
lerastarynets, line 3: Invalid value 'space' for 'justify-content'. It should be 'space-between', 'space-around' or 'space-evenly'

src/styles/blocks/controls.scss
```diff
@@ -0,0 +1,21 @@
...
+  align-items: center;
+  flex-flow: column wrap;
```
lerastarynets, line 5: The 'flex-flow' property is not necessary here because the default direction of 'flex-direction' is 'row' and the default value of 'flex-wrap' is 'nowrap'. This means that items will arrange in a row and not wrap to the next line by default. You can remove this line of code.

src/components/Carousel.scss
```diff
@@ -1 +1,35 @@
...
+    transform: translateX(var(--transform-offset));
+    transition: transform var(--animation-duration) ease-out;
+  }
+
+  &__image {
+    width: var(--image-size);
+    height: var(--image-size);
+  }
+
+  & li:not(:last-child) img {
```
lerastarynets, line 32: Avoid using tag selectors with class selectors. It’s less specific and can lead to unwanted side effects in your CSS. Maybe create a new class for img inside li elements.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,275 @@
...
+                const val = parseInt(e.target.value, 10);
+
+                if (val < 1) {
+                  setInputFrameSize('1');
+                } else if (val > images.length) {
+                  setInputFrameSize(images.length.toString());
+                } else {
+                  setInputFrameSize(e.target.value);
+                }
+              }}
```
lerastarynets, line 259: Pls move the function inside `onChange` handler to a separate const and use as a callback here

src/components/Carousel.tsx
```diff
@@ -1,18 +1,275 @@
...
+              value={inputValue}
+              min="100"
+              max="350"
+              onChange={(e) => {
+                const newValue = parseInt(e.target.value, 10);
+
+                if (!Number.isNaN(newValue)) {
+                  setInputValue(newValue);
+                }
+              }}
```
lerastarynets, line 183: Pls move the function inside `onChange` handler to a separate const and use as a callback here

src/components/Carousel.tsx
```diff
@@ -1,18 +1,275 @@
...
-      <li><img src="./img/1.png" alt="3" /></li>
-      <li><img src="./img/1.png" alt="4" /></li>
-    </ul>
-
-    <button type="button">Prev</button>
-    <button type="button">Next</button>
-  </div>
-);
+/* eslint-disable no-alert */
+/* eslint-disable no-console */
```
lerastarynets, line 2: Avoid disabling eslint rules

src/components/Carousel.tsx
```diff
@@ -93,8 +98,17 @@ const Carousel: React.FC<CarouselProps> = ({
...
   };
 
-  // ???????????
-  // type checking needed
+  const handleStepChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const val = parseInt(e.target.value, 10);
+
+    if (val < 1) {
+      setInputStep('1');
+    } else if (val > images.length) {
```
anastasiia-tilikina, line 106: try to avoid `else if` structure, it look a bit unclear.
use `return`

```javascript
if (val < 1) {
  setInputStep('1');
  
  return
}

if (// your code

```

src/components/Carousel.tsx
```diff
@@ -108,8 +122,19 @@ const Carousel: React.FC<CarouselProps> = ({
     }
   };
 
+  const handleFrameSizeChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const val = parseInt(e.target.value, 10);
+
+    if (val < 1) {
+      setInputFrameSize('1');
+    } else if (val > images.length) {
```
anastasiia-tilikina, line 130: same here

src/components/Carousel.tsx
```diff
@@ -1,18 +1,275 @@
...
+    } else {
+      alert(
+        'Invalid frame size. Must be between 1 and the total number of images.',
+      );
+    }
+  };
+
+  useEffect(() => {
+    // console.log('useEffect has been triggered!');
+    const scrollWidth = offset * (dynamicItemWidth + 10);
```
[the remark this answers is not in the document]
anastasiia-tilikina: not fixed

src/App.scss
```diff
@@ -1 +1,13 @@
...
+.App {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  text-align: center;
+  min-height: 50vh;
+  justify-content: center;
+
+  &__title {
+    text-align: center;
```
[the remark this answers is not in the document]
anastasiia-tilikina: not fixed

src/App.tsx
anastasiia-tilikina: not fixed

anastasiia-tilikina (changes requested):
I see you haven't fixed issues from the previous review. 
please make sure that everything is done before re-requesting it again.

also, I'd suggest you extract the form to a separate component and move states to `App`.
now your `Carousel` props are useless because you replace those values with local state.

Anastasiia-Svintsova (changes requested):
- Please, make sure you fixed all previous comments 
- Calculate the scroll width correctly 
image
image

src/App.scss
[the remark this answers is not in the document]
anastasiia-tilikina: still not fixed

src/components/Carousel.tsx
anastasiia-tilikina: not fixed

src/components/Carousel.tsx
```diff
@@ -1,18 +1,295 @@
...
+
+  const [dynamicAnimationDuration, setDynamicAnimationDuration]
+    = useState<number>(animationDuration);
+
+  const [inputStep, setInputStep] = useState<string>(step.toString());
+  const [localStep, setLocalStep] = useState<number>(step);
+
+  const [inputFrameSize, setInputFrameSize]
+    = useState<string>(frameSize.toString());
+  const [localFrameSize, setLocalFrameSize] = useState<number>(frameSize);
```
anastasiia-tilikina, line 38: please look at this code. you're putting your props as initial values to your local state and then you change them locally. this makes no sense) 

move the logic of changing state to the App component.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,295 @@
...
+
+    document.documentElement.style.setProperty('--transform-offset', `-${scrollWidth}px`);
+    document.documentElement.style.setProperty('--image-size', `${dynamicItemWidth}px`);
+    document.documentElement.style.setProperty('--frame-size', `${localFrameSize}`);
+    document.documentElement.style.setProperty(
+      '--animation-duration', `${animationDuration}ms`,
+    );
+    document.documentElement.style.setProperty(
+      '--animation-duration', `${dynamicAnimationDuration}ms`,
+    );
```
anastasiia-tilikina, line 161: This part is still not fixed as well.
Interacting with DOM directly in React is bad practice.
We need to avoid it.
Consider adding some inline styles to your elements.

anastasiia-tilikina (changes requested):
Great! you've figured it out with your transform calculations! 
One step at a time)

let's look at other issues that are still not fixed!
Please read again my previous comments.

If it is still not clear consider signing up for Q&A

src/components/Carousel.tsx
```diff
@@ -1,18 +1,245 @@
...
+
+  useEffect(() => {
+    const scrollWidth = offset * (itemWidth);
+
+    document.documentElement.style.setProperty('--transform-offset', `-${scrollWidth}px`);
+    document.documentElement.style.setProperty('--image-size', `${itemWidth}px`);
+    document.documentElement.style.setProperty('--frame-size', `${frameSize}`);
+    document.documentElement.style.setProperty(
+      '--animation-duration', `${animationDuration}ms`,
+    );
```
lerastarynets, line 132: Do not manipulate DOM directly in react

src/components/Carousel.tsx
```diff
@@ -1,18 +1,245 @@
...
+  }, [offset, itemWidth, frameSize, animationDuration]);
+
+  return (
+    <>
```
lerastarynets, line 136: Looks like this fragment is redundant

lerastarynets (changes requested):
Prev comments should be fixed
Pls ask for help in **fe_chat** if you don't understand something

src/components/Carousel.tsx
domkab: image

i think its not

Anastasiia-Svintsova (approved):

PR #480, opened 2023-08-11
Solution

G37ua:
[DEMO LINK](https://g37ua.github.io/react_carousel/)


Anastasiia-Svintsova (changes requested):
- Hey! Nice design. But I can't review the project because pictures are not displayed. 
image

G37ua:
I put an extra dot in the path to the picture. On the local server, everything was fine, but on the github, the pictures disappeared

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,77 @@
...
+          disabled={currentIndex >= images.length - frameSize && !infinite}
+          data-cy="next-button"
+        />
+      </div>
+    </div>
+  );
+};
+
+// Define default props
+Carousel.defaultProps = {
```
Anastasiia-Svintsova, line 71: No need to repeat it twice. You've already had it 
image

Anastasiia-Svintsova (changes requested):
Keep going! Test it more thoroughly before requesting a new review, please 

- Make buttons disabled if the user can't scroll anymore 
- There shouldn't be empty spaces. Calculate the scroll width correctly
image
 
- Would be great to add min and max values for inputs 
image

https://github.com/mate-academy/react_carousel/assets/99403198/842492d2-4b27-4f09-926b-2b7979252c22



- The frame size value shouldn't be bigger than the amount of the image 
image

sTorba24 (approved):
Well done! Pay attention about inifinite mode, it should work in both sides :) Now it works only when i click right button:)
image

PR #481, opened 2023-08-11
add task solution

NadiiaR:
[DEMO LINK](https://NadiiaR.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,7 +24,7 @@ const App: React.FC = () => {
 
   const { images } = state;
   const [sizeImages, setSizeImages] = useState(130);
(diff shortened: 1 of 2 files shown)


sTorba24 (changes requested):
Good work! Lets improve it

default step should be 3
image

check how carousel works on this value, i can't get to the last 2 elements

Also, infinity mode has the same problem


then, i see 2 emojis and 1 blank place(it shouldn't be so)
image
If not enough elements to scroll as expected(with step 3, for example, but have only 2 left, so scroll only 2 to the end)

sTorba24 (approved):

PR #484, opened 2023-08-14
Develop

Luk2asz:
[DEMO LINK](https://Luk2asz.github.io/react_carousel/)


src/App.tsx
```diff
@@ -20,18 +26,170 @@ class App extends React.Component<{}, State> {
...
+        animationDuration: this.getInputs('animationDuration'),
+        frameSize: this.getInputs('frameSize'),
+        itemWidth: this.getInputs('itemWidth'),
+        moveLeft: left,
+      };
+    });
+  };
+
+  nextButton = () => {
+    this.setState((state) => {
```
darokrk, line 68: same here as above

src/App.tsx
```diff
@@ -20,18 +26,170 @@ class App extends React.Component<{}, State> {
...
+        </div>
+
+        <form action="" className="form">
+          <label htmlFor="itemWidth">
+            item width:
+            <input
+              type="number"
+              id="itemWidth"
+              name="itemWidth"
+              defaultValue="130"
```
darokrk, line 156: same here as I mentioned before, please use the controlled input component by react, so you should onChange() method for every input to get values from them

src/App.tsx
```diff
@@ -20,18 +26,170 @@ class App extends React.Component<{}, State> {
...
+            />
+          </label>
+
+          <br />
```
darokrk, line 160: please remove these `<br />` tags they are redundant, if you want to break the lines you can do this by CSS styling like display: inline-block or block or just wrap elements into <div>. <br /> should be used only on specific scenarios

src/components/Carousel.scss
```diff
@@ -1 +1,40 @@
-// style go here
+.Carousel {
```
darokrk, line 1: rename class to lowercase name

darokrk (changes requested):
Good job, UI and carousel works 👏🏽  but few mentions added for code improvements 👨🏽‍💻 🛠️

src/App.scss
```diff
@@ -1 +1,14 @@
...
+  text-align: right;
+}
+
+.input {
+  margin-top: 10px;
+  width: 100px;
+  display: inline-block;
+}
+
+.div {
```
darokrk, line 12: please rename it to a more descriptive name, like `.container` or `.wrapper` the `.div` is to more generic name and it's not a good practice to name classes like that

src/components/Carousel.tsx
```diff
@@ -1,18 +1,55 @@
...
+      frameSize,
+      itemWidth,
+      moveLeft,
+    } = this.props;
+
+    return (
+      <ul
+        className="carousel__list"
+        style={{
+          display: 'flex',
```
darokrk, line 27: These attributes like `display: flex` and `overflow: hidden` can be moved into `.carousel__list` css file

src/components/Carousel.tsx
```diff
@@ -1,18 +1,55 @@
...
+          width: `${frameSize * (itemWidth + 10)}px`,
+          overflow: 'hidden',
+          transition: `all ${animationDuration}ms ease-out`,
+        }}
+      >
+        {images.map((image, index) => (
+          <li
+            key={image}
+            style={{ listStyleType: 'none' }}
```
darokrk, line 36: same for that, please move this style to CSS file, bcs it static no dynamic style

darokrk (changes requested):
Almost ready 👨🏽‍💻  left last mentions for fix and we got this 🛠️ 💪🏽

PR #485, opened 2023-08-16
done

maxonshev:
https://maxonshev.github.io/react_carousel/

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -55,6 +55,7 @@ const Carousel: React.FC<Props> = ({
             <img
               src={image}
               alt={index.toString()}
+              className="Carousel__image"
               style={{ width: itemWidth }}
             />
           </li>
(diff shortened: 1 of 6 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,87 @@
...
+        </button>
+
+        <button
+          type="button"
+          className={cn('button', {
+            'button--disabled': offset >= images.length - frameSize
+              && !infinite,
+          })}
+          onClick={handleNextClick}
+          disabled={offset >= images.length - frameSize && !infinite}
```
etojeDenys, line 76: move this condition to some variable for better readability

src/components/Carousel.tsx
```diff
@@ -1,18 +1,87 @@
...
+          data-cy="prev"
+        >
+          &lt;
+        </button>
+
+        <button
+          type="button"
+          className={cn('button', {
+            'button--disabled': offset >= images.length - frameSize
+              && !infinite,
```
etojeDenys, line 73: and reuse everywhere

etojeDenys (changes requested):
image
let's add some min and max values to the inputs

etojeDenys (approved):
good job

PR #494, opened 2023-08-29
Solution for carouselle

AndrzejPec:
[DEMO LINK](https://andrzejpec.github.io/react_carousel/)


src/App.scss
```diff
@@ -1 +1,10 @@
-// style go here
+.carousel {
```
klaudia-klis, line 1: I don't see this class in your code, are you using it somewhere?

klaudia-klis (approved):
Code looks good to me 👍🏼

src/components/Carousel.tsx
```diff
@@ -1,18 +1,97 @@
...
+
+export const Carousel: React.FC<Props> = ({
+  images,
+  imageWidth,
+  frameSize,
+  step,
+  animationDuration,
+  infinite = false,
+}) => {
+  const gap = 25;
```
choeqq, line 22: this name should be uppercased - also please try not to keep constants inside of components

src/components/Carousel.tsx
```diff
@@ -1,18 +1,97 @@
...
+  const containerWidth = (imageWidth + gap) * frameSize;
+  const totalWidth = (imageWidth + gap) * images.length;
+  const maxOffset = -(totalWidth - containerWidth);
+
+  const [currentOffset, setCurrentOffset] = useState(0);
+  const moveLeft = () => {
```
choeqq, line 28:
```suggestion
  const [currentOffset, setCurrentOffset] = useState(0);
  
  const moveLeft = () => {
```

choeqq (changes requested):

src/App.scss
AndrzejPec: good point ,thx

asakevych (changes requested):
Great job!! 🚀  it works, but it would be great if you provide small changes:
1. arrows scroll to opposite side (right arrow scrolls left)
2.  you can make your layout look better 
  - add some styles to form (at least increase spaces between fields) 
  - center your arrows and make them  in one line, notone above the other
3. please add some radio button to your form to check if scroll is infinite or not (it seems it should be configurable too but not leased in readme, so it is nice to have but if you don't want to do it, so it can stay as it is now)

asakevych (changes requested):
1. tests are not passed 
2. arrows still works to opposite side 
3. infite doesn't work -> button is disabled when no more image aavailable, but it should be enabled and scroll to the beginnig

image

choeqq (approved):
All errors seem to be fixed, well done ✅

PR #502, opened 2023-09-03
solution

Absolution26:
[DEMO LINK](https://Absolution26.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.scss
+++ b/src/components/Carousel.scss
@@ -1,1 +1,9 @@
-// style go here
+.Carousel {
+  &__list {
+    border: 3px solid black;
+    padding: 0;
+    display: flex;
+    list-style: none;
+    overflow: hidden;
+  }
+}
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,147 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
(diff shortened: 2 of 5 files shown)


src/components/Carousel.scss
```diff
@@ -1 +1,9 @@
-// style go here
+.Carousel {
+  &__list {
+    border: 3px solid black;
```
OleziO, line 3: Use HEX instead of color names.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,147 @@
...
+                width: `${smileWidth}px`,
+                transform: `translate(${currentSmile}px)`,
+                transition: `all ${animationTime / 1000}s ease-in`,
+              }}
+            />
+          </li>
+        ))}
+      </ul>
+
+      Width:&nbsp;
```
OleziO, line 76: Wrap it with label.

OleziO: Check other similar cases.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,147 @@
...
+      Animation duration:&nbsp;
+      <input
+        type="number"
+        min="1000"
+        max="5000"
+        value={`${animationTime}`}
+        onChange={event => setAnimationTime(+event.target.value)}
+      />
+
+      {/* sorry, Could not beautifully implement an infinite carousel */}
```
OleziO, line 119: Remove all comments from the code.

OleziO (changes requested):

maxim2310 (approved):
Good work!!!
why do you add infinite checkbox if you didnt realize this func 
image

PR #504, opened 2023-09-04
Develop

dimarogkov:
[DEMO LINK](https://dimarogkov.github.io/react_carousel/)


src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,136 @@
...
+  step,
+  animationDuration,
+}) => {
+  const [currentSlide, setCurrentSlide] = useState(0);
+  const [isPrevBtnDisabled, setPrevBtnDisabled] = useState(true);
+  const [isNextBtnDisabled, setNextBtnDisabled] = useState(false);
+  const [imgWidth, setImgWidth] = useState(itemWidth);
+  const [imgCount, setImgCount] = useState(frameSize);
+  const [carouselStep, setCarouselStep] = useState(step);
+  const [duration, setDuration] = useState(animationDuration);
```
VitaliyBondarenko1982, line 34:
```suggestion
  const [currentSlide, setCurrentSlide] = useState(0);
```
enough to keep in state currentSlide.
isPrevBtnDisabled and isNextBtnDisabled will be change based on currentSlide.

imgWidth, imgCount, carouselStep duration already in parent component state, no need to duplicate them here in state.

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,136 @@
...
+        && currentSlide <= endOfList
+      ) {
+        setCurrentSlide((imgWidth * (images.length - imgCount) * -1));
+        setNextBtnDisabled(true);
+      }
+    };
+
+    btn.classList.contains('Carousel__btn--prev')
+      ? slideBack()
+      : slideForward();
```
VitaliyBondarenko1982, line 88: make two different handlers for back and forward buttons. no need to check classname. if classname will change this logic also will be broken

VitaliyBondarenko1982 (changes requested):
Good job.
Needs some improvements.

- view is broken for itemWidth approximately 100
Screenshot from 2023-09-04 17-21-09
- check your logic. after change item width should also be visible whole number of images.
not part of first and part of last.
chrome-capture-2023-8-4

src/components/Options/Options.tsx
```diff
@@ -0,0 +1,77 @@
...
+import './Options.scss';
+
+interface Props {
+  options: [string, number, React.Dispatch<React.SetStateAction<number>>][];
+  setCurrentSlide: React.Dispatch<React.SetStateAction<number>>,
+  imagesLength: number;
+}
+
+export const Options: React.FC<Props> = ({
+  options,
```
etojeDenys, line 11: if you want to make your component more reusable and readable let's change the options a little bit:

i suggest you to pass the array of objects like this
```
[{
  name: '',
  value: '',
  set: () => {},

  max: ''
  min: '', // let's add these fields it will remove the extra logic from the options component 
}]
```

etojeDenys: so you need to create the interface for option object

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,131 @@
...
+        >
+          Prev
+        </button>
+        <button
+          className={cn(
+            'Carousel__btn',
+            'Carousel__btn--next',
+            {
+              'Carousel__btn--disabled': currentSlide === imgWidth
+                * (images.length - imgCount) * -1,
```
etojeDenys, line 113: let's move this difficult condition to some var

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,131 @@
...
+  };
+
+  const slideBack = () => {
+    if (currentSlide >= imgWidth * carouselStep * -1) {
+      setCurrentSlide(0);
+    }
+
+    if (
+      currentSlide >= imgWidth * images.length * -1
+      && currentSlide <= beginOfList
```
etojeDenys, line 60: we can move it to some variable
`currentSlide >= imgWidth * images.length * -1`

etojeDenys (changes requested):
feel free to ask in the chat if you have some questions

IvanFesenko (approved):

PR #513, opened 2023-09-09
Add task solution

NikitaOrliak:
[DEMO LINK](https://NikitaOrliak.github.io/react_carousel/)

To be honest, my tests just don't work. I used the live server to check the reports and found problems with the pictures. But as I see, other people have the same problems with the tests in this task, so I want to try to publish task without npm tests ;)

Thank you!

### DIFF
--- a/src/components/Carousel.scss
+++ b/src/components/Carousel.scss
@@ -1,1 +1,22 @@
 // style go here
+.Carousel {
+  &__container {
+    overflow: hidden;
+  }
+
+  &__list {
+    display: inline-flex;
+    list-style-type: none;
+    padding: 0;
+  }
+
+  &__params {
+    display: flex;
+    flex-direction: column;
+    width: fit-content;
+
+    &--label {
+      margin-top: 15px;
+    }
+  }
+}
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,151 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
-      <li><img src="./img/1.png" alt="3" /></li>
-      <li><img src="./img/1.png" alt="4" /></li>
-    </ul>
+interface Props {
+  images: string[];
+}
 
-    <button type="button">Prev</button>
-    <button type="button">Next</button>
-  </div>
-);
+const Carousel: React.FC<Props> = ({ images }) => {
+  const [translate, setTranslate] = useState(0);
+  const [itemWidth, setItemWidth] = useState(130);
+  const [frameSize, setFrameSize] = useState(3);
+  const [step, setStep] = useState(3);
+  const [animationDuration, setAnimationDuration] = useState(1000);
+
+  const maxStep = 100 - ((100 / images.length) * frameSize);
+  const currentStep = (100 / images.length) * step;
+
+  function nextClick() {
+    setTranslate(prevTranslate => {
+      return prevTranslate + currentStep > maxStep
+        ? maxStep
+        : prevTranslate + currentStep;
+    });
+  }
+
+  function prevClick() {
+    setTranslate(prevTranslate => {
+      return prevTranslate - currentStep < 0
+        ? 0
+        : prevTranslate - currentStep;
+    });
+  }
+
+  function isFirstImage() {
+    return translate <= 0;
+  }
+
+  function isLastImage() {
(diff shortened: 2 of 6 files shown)


src/components/Carousel.scss
```diff
@@ -1 +1,22 @@
 // style go here
```
OleziO, line 1: Remove all comments from the code.

OleziO (changes requested):
Add infinity mode.
image

src/components/Carousel.tsx
```diff
@@ -1,18 +1,186 @@
...
+
+      return prevTranslate - currentStep < 0
+        ? 0
+        : prevTranslate - currentStep;
+    });
+  }
+
+  function isFirstImage() {
+    return translate <= 0;
+  }
```
VitaliyBondarenko1982, line 45:
```suggestion
  const isFirstImage = translate <= 0;
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,186 @@
...
+
+  function isLastImage() {
+    return translate >= maxStep;
+  }
```
VitaliyBondarenko1982, line 49:
```suggestion
  const isLastImage =   translate >= maxStep;
```
looks like redundant to make function in these cases.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,186 @@
...
+          className="Carousel__params--label"
+        >
+          Width of image in px:
+        </label>
+
+        <input
+          type="number"
+          id="itemId"
+          value={itemWidth}
+          onChange={value => setItemWidth(Number(value.target.value))}
```
VitaliyBondarenko1982, line 116:
```suggestion
          onChange={event => setItemWidth(Number(event.target.value))}
``` 
common name of first argument for event handler is `event` or something like that

src/components/Carousel.tsx
```diff
@@ -1,18 +1,186 @@
...
+  images: string[];
+}
+
+const Carousel: React.FC<Props> = ({ images }) => {
+  const [translate, setTranslate] = useState(0);
+  const [itemWidth, setItemWidth] = useState(130);
+  const [frameSize, setFrameSize] = useState(3);
+  const [step, setStep] = useState(3);
+  const [animationDuration, setAnimationDuration] = useState(1000);
+  const [infinityMode, setInfinityMode] = useState(false);
```
VitaliyBondarenko1982, line 14: suggest to make one state for all inputs as object and use event.target.name for change proper field

```suggestion
const [inputs, setInputs] = useState({
  itemWidth: 0,
  ...
});
```
also caruousel shouldn't to know about inputs and except just images and values of inputs.

Screenshot from 2023-09-09 16-16-42

move inputs and state in App component or create separate component, but it should be in App component.

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some improvements.

NikitaOrliak:
Honestly, I have 2 problems:
1. I don't know how to change the state of an object if I use something like const [inputs, setInputs] = useState({ ... }). I can read properties, but don't know how to change them.
2. I don't need post and infinite in Carousel. By convention I MUST take them, but in the Carousel.jsx file I don't use them , then lint cusses at me. Also, I need the translate param to change translate: transform { translate.param }.
Hopefully I can turn a blind eye on these params :)

Thanks for your posts!

TarasHoliuk:
1. You may create 1 function to update each input (except the checkbox) when the state is an object with all fields:

```
function handleChange(event) {
  const { name, value } = event.target;
  
  setInputs(inputs => ({
    ...inputs,
    [name]: Number(value),
  }))
}
```

2. > I don't need post and infinite in Carousel.

IMO it's ok, since you separate the actually Carousel rendering from the controls, so you can keep it as it is

src/components/Functional/Functional.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  return (
+    <>
+      <Carousel
+        images={images}
+        // step={step}
+        frameSize={frameSize}
+        itemWidth={itemWidth}
+        animationDuration={animationDuration}
+        // infinite={infinityMode}
```
TarasHoliuk, line 60: delete comments

TarasHoliuk (changes requested):

src/components/Functional/Functional.tsx
```diff
@@ -0,0 +1,180 @@
...
+            Duration of animation in ms:
+          </label>
+
+          <input
+            type="number"
+            id="durationAnimationId"
+            value={inputs.animationDuration}
+            min={0}
+            name="animationDuration"
+            onChange={event => handleChange(event)}
```
etojeDenys, line 158:
```suggestion
            onChange={handleChange}
```
we can simplify it

etojeDenys (approved):
lgtm

PR #518, opened 2023-09-14
add task solution

andrew17752985:
https://andrew17752985.github.io/react_carousel/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -32,7 +32,7 @@ const App: React.FC = () => {
         <label>Item Width: </label>
         <input
           type="number"
-          value={itemWidth.toString()} // Конвертуємо число у рядок
+          value={itemWidth.toString()}
           onChange={(e) => setItemWidth(parseInt(e.target.value, 10))}
         />
       </div>
@@ -41,7 +41,7 @@ const App: React.FC = () => {
         <label>Frame Size: </label>
         <input
           type="number"
-          value={frameSize.toString()} // Конвертуємо число у рядок
+          value={frameSize.toString()}
           onChange={(e) => setFrameSize(parseInt(e.target.value, 10))}
         />
       </div>
@@ -50,7 +50,7 @@ const App: React.FC = () => {
         <label>Step: </label>
         <input
           type="number"
-          value={step.toString()} // Конвертуємо число у рядок
+          value={step.toString()}
           onChange={(e) => setStep(parseInt(e.target.value, 10))}
         />
       </div>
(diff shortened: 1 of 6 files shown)


volodymyr-soltys97 (changes requested):
The task is not completely completed and it is not working
image

src/App.tsx
```diff
@@ -1,39 +1,82 @@
...
+  const [animationDuration, setAnimationDuration] = useState<number>(1000);
+
+  return (
+    <div className="App">
+      <h1>
+        Carousel with
+        {' '}
+        {images.length}
+        {' '}
+        images
```
TarasHoliuk, line 32: use string template in such cases

src/components/Carousel.scss
```diff
@@ -1 +1,54 @@
...
+button[data-cy="prevButton"] {
+  left: 10px;
+}
+
+button[data-cy="nextButton"] {
+  right: 10px;
+}
+
+button.disabled {
+  cursor: not-allowed;
```
TarasHoliuk, line 52: in my case it just disappeared - just show common cursor

src/components/Carousel.scss
```diff
@@ -1 +1,54 @@
...
+  margin-right: 10px;
+}
+
+
+.Carousel__list img {
+  width: 130px;
+  height: 130px;
+}
+
+button {
```
TarasHoliuk, line 29: do not add styles with tag selector

src/components/Carousel.scss
```diff
@@ -1 +1,54 @@
...
+  height: 40px;
+  color: #4e4949;
+  border-radius: 50%;
+}
+
+button[data-cy="prevButton"] {
+  left: 10px;
+}
+
+button[data-cy="nextButton"] {
```
TarasHoliuk, line 47: add corresponding class instead of using attribute selector

src/components/Carousel.tsx
```diff
@@ -1,18 +1,88 @@
...
+    }
+  };
+
+  useEffect(() => {
+    const interval = setInterval(handleNextClick, animationDuration);
+
+    return () => clearInterval(interval);
+  }, [currentIndex, handleNextClick, animationDuration]);
+
+  const visibleImages = images.slice(currentIndex, currentIndex + frameSize);
```
TarasHoliuk, line 54: in case frame size is 3 you should always show 3 images (last 3 in this case):
image

src/components/Carousel.tsx
```diff
@@ -1,18 +1,88 @@
...
+  animationDuration?: number;
+  infinite?: boolean;
+}
+
+const Carousel: React.FC<CarouselProps> = ({
+  images,
+  itemWidth = 130,
+  frameSize = 3,
+  step = 3,
+  animationDuration = 1000,
```
TarasHoliuk, line 19: you should use it to change images smoothly on buttons click

TarasHoliuk (changes requested):
If you have more questions feel free to ask in the chat

volodymyr-soltys97 (changes requested):
The slider now automatically switches to the last image, it needs to switch only when you click on the buttons, also try to add a smooth transition between images
image

PR #519, opened 2023-09-15
solution

Spec0s:
[Carousel](https://spec0s.github.io/fe-program/js/extra/react-typescript)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
       infinite={false}
     />
     ```
-    
+
 ## REQUIREMENTS:
 
 1. The title of the page should contain "Carousel"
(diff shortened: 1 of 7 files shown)


Anastasiia-Svintsova (changes requested):
- Add min value for the size 
image

- The frame size can't be bigger than the carousel items amount. 
- Add the max-width for the carousel. Now I can't see buttons if the carousel size is bigger than the screen width
image

- Fix linter errors. Don't use any for types

Spec0s:
please need a hint on types, cant figure it out and maybe have good source to read about it

lerastarynets (approved):
lgtm 🚀

PR #524, opened 2023-09-18
Solution

KarolinaWiktoria:
[DEMO LINK](https://KarolinaWiktoria.github.io/react_carousel/)


src/components/LabeledInput.tsx
```diff
@@ -0,0 +1,34 @@
...
+type InputProps = {
+  name: string;
+  label:string;
+  value: string | number;
+  type?: string;
+  min?: number;
+  max?: number;
+  checked?: boolean;
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => void
+};
```
etojeDenys, line 13:
```suggestion
type InputProps = {
  label:string;
} & React.HTMLProps<HTMLInputElement>;
```

we can just expand the input type element ([link](https://stackoverflow.com/questions/62180651/how-to-extend-jsx-types-for-typescript))

etojeDenys (changes requested):

DarkMistyRoom (changes requested):
Good job! Only one suggestion:
when the Infinite option is on move buttons swipe the carousel to the end ignoring the step size. Fix it. Infinity should only make the carousel cyclic

lerastarynets (changes requested):
prev comments fixed but next button doesn't seem to be working in infinite mode

https://github.com/mate-academy/react_carousel/assets/77551820/7f510864-b998-455b-a490-6cae3ebdcf0f

VitaliyBondarenko1982 (changes requested):
Looks like infinite mode still not work.
 Or try to redeploy your solution.
chrome-capture-2023-8-19

etojeDenys (approved):
good job

PR #525, opened 2023-09-19
Develop

Sirius9312:
DEMO LINK: https://Sirius9312.github.io/react_carousel/
COMMIT Carousel Task2

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,6 +4,11 @@ import Carousel from './components/Carousel';
 
 interface State {
   images: string[];
+  step: number,
+  frameSize: number,
+  itemWidth: number,
+  animationDuration: number,
+  infinity: boolean,
 }
 
 class App extends React.Component<{}, State> {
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -20,17 +25,103 @@ class App extends React.Component<{}, State> {
...
+        <div className="App__inputs-container">
+          <label>
+            Step:
+            <input
+              type="number"
+              min="1"
+              max="10"
+              value={step}
+              step="1"
+              onChange={(e) => this.setState({ step: +e.target.value })}
```
sTorba24, line 72: create common function for setting state

handleChange = (e) => setState... 
 
here you can get also input name(before set it to the same value as name of state field)

src/components/Carousel.tsx
```diff
@@ -1,18 +1,112 @@
...
+                />
+              </li>
+            ))
+          }
+        </ul>
+
+        <div className="Carousel__buttons-container">
+          <button
+            type="button"
+            className={movingCarousel === 0 && !infinite
```
sTorba24, line 87:
```suggestion
            className={!movingCarousel && !infinite
```

sTorba24 (changes requested):

anastasiia-tilikina (approved):
GJ

PR #526, opened 2023-09-26
React__Carousel

levchenko-dmytro:
[DEMO LINK](https://levchenko-dmytro.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,6 +4,11 @@ import Carousel from './components/Carousel';
 
 interface State {
   images: string[];
+  step: number,
+  frameSize: number,
+  itemWidth: number;
+  animationDuration: number;
+  infinite: boolean;
 }
 
 class App extends React.Component<{}, State> {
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -20,17 +25,121 @@ class App extends React.Component<{}, State> {
...
+          <label htmlFor="stepId">
+            Step:
+            <input
+              id="stepId"
+              className="inputs__input"
+              type="number"
+              value={step}
+              min={1}
+              max={5}
+              step={1}
```
lerastarynets, line 88: consider moving all such numbers to consts with consistent names, avoid usong magic numbers

src/components/Carousel.tsx
```diff
@@ -1,18 +1,104 @@
...
+        width: `${frameSize * itemWidth}px`,
+        transition: `width ${animationDuration}ms`,
+      }}
+    >
+      <ul
+        className="Carousel__list"
+      >
+        {images.map(image => (
+          <li
+            key={images.indexOf(image) + 1}
```
lerastarynets, line 62: it's not a good practice, to assign key prop a value corresponding to index of item in array
consider turning your `images` from arr of strings to arr of objects of type {id: , url: } and use `id` value of every object for key

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

volodymyr-soltys97 (approved):
Great work 🔥

PR #530, opened 2023-09-29
Implement_image_carousel

rialleons:
- [DEMO LINK](https://rialleons.github.io/react_carousel/)

Unfortunately I couldn't make autoscroll start on button click. It's either scrolls just once or don't listen to button at all. When I tried to start whole Carousel component on button click then tests became failed.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,113 @@
-import React from 'react';
+import React, { useState } from 'react';
+
 import './App.scss';
-import Carousel from './components/Carousel';
+import './services/reset.scss';
+import { Carousel } from './components/Carousel';
+import { CarouselParameters } from './types/CarouselParameters';
+import { getNumbers } from './services/utils';
 
-interface State {
-  images: string[];
-}
+const defaultCarouselParameters:CarouselParameters = {
+  step: 3,
+  frameSize: 3,
+  itemWidth: 130,
+  animationDuration: 1000,
+};
 
-class App extends React.Component<{}, State> {
-  state = {
-    images: [
-      './img/1.png',
-      './img/2.png',
-      './img/3.png',
-      './img/4.png',
-      './img/5.png',
-      './img/6.png',
-      './img/7.png',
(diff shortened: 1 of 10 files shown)


sTorba24 (changes requested):
Good work!

Text your fe_chat and try to find solution with mentors 

It'll be not comfortable to solve it here, because mentors can see your comment only when task is in check-queue.
I'm sure you can deal with your problem, keep it up :)

src/App.tsx
```diff
@@ -1,39 +1,99 @@
...
+  );
+
+  const images = getImagesUrls;
+  const isInfinite = false;
+
+  const handleFieldChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setCarouselParameters({
+      ...carouselParameters,
+      [event.target.name]: +event.target.value,
+    });
```
VitaliyBondarenko1982, line 31:
```suggestion
   const { name, value} = event.target;
   
    setCarouselParameters({
      ...carouselParameters,
      [name]: +value,
    });
```
you can ditrructure event.target

src/components/Carousel.tsx
```diff
@@ -1,18 +1,141 @@
...
+      <ul className="Carousel__list">
+        {rotatedImages.map((image, index) => {
+          const isShown = (index < frameSize);
+
+          return (
+            <li
+              key={image}
+              className="Carousel__item"
+              style={{
+                display: isShown ? 'block' : 'none',
```
VitaliyBondarenko1982, line 96: make change visible images smoothly.
for it instead display set  to style `transform: translateX()` and transition with  transition-duration animationDuration.
make state for example `position`, and change it after click on carousel buttons.
it will be value for translateX and will count base on step and itemWidth.
if you will have any questions - feel_free to ask in fe_chat

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):
Well done!

PR #535, opened 2023-10-04
'Working'

Hanna-Balabukha:
https://Hanna-Balabukha.github.io/react_carousel/


sTorba24 (changes requested):
Well done! Add infinity mode

PR #549, opened 2023-10-18
Develop

Vasyl-Zhyliakov:
[DEMO LINK](https://Vasyl-Zhyliakov.github.io/react_carousel/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
       infinite={false}
     />
     ```
-    
+
 ## REQUIREMENTS:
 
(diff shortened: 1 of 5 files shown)


maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

1. dont use transition for background size 
image
2. but transition dont work when i click next or prev 
3. infinity mode is not working

TarasHoliuk (approved):

PR #556, opened 2023-10-22
add task solution

4xmplme:
[DEMO LINK](https://4xmplme.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,157 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
-import Carousel from './components/Carousel';
+import { Carousel } from './components/Carousel';
 
-interface State {
-  images: string[];
-}
+const images = [
+  './img/1.png',
+  './img/2.png',
+  './img/3.png',
+  './img/4.png',
+  './img/5.png',
+  './img/6.png',
+  './img/7.png',
+  './img/8.png',
+  './img/9.png',
+  './img/10.png',
+];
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,39 +1,157 @@
...
+  './img/5.png',
+  './img/6.png',
+  './img/7.png',
+  './img/8.png',
+  './img/9.png',
+  './img/10.png',
+];
+
+const App: React.FC = () => {
+  const [corouselSettings, setCorouselSettings] = useState({
```
volodymyr-soltys97, line 19:
```suggestion
  const [carouselSettings, setCarouselSettings] = useState({
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,104 @@
...
+  animationDuration,
+  infinite,
+  currentImage,
+  setCurrentImage,
+}) => {
+  const carouselWidth = itemWidth * frameSize;
+  const translateDistance = currentImage * itemWidth;
+  const maxImages = images.length - frameSize;
+
+  function handlePrev() {
```
volodymyr-soltys97, line 30: Consider will be better if you use a function expression here
```suggestion
  const handlePrev = () => {
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve something

volodymyr-soltys97 (approved):
Awesome 🔥

PR #559, opened 2023-10-29
implementation

Daniilart01:
[DEMO LINK](https://Daniilart01.github.io/react_carousel/)
I have a trouble with test, everything works but tests failed


volodymyr-soltys97 (changes requested):
Good job 👍 
But now it is not possible to change the values of these fields, it needs to be fixed so that there is such an opportunity
image

Daniilart01:
We have this note that suggests u to change values using keyboard arrows. This allows us to restrict any inputs which a different of [1-9]
image

Daniilart01:
Because of our state updates with every inputed character for better expirience, and input always displays current state `value={currentStateStep}` it can not be allowed to set empty line (in case you want to delete `1` from input and type `9`). Because of this trouble, u can use arrows to manipulate data in input or highlight number and type another one. 
There is one more option, we can remove limits for this fields and in this way, we can allow user to set manualyy all fields, but in moment when field will be empty - carousel will be probably crashed.
In first implementation all of fields was able to set values via keyboard with value-step 1 and for animation step was 100, but because of cypress that require to change value manualy, I had to change this logic and allow users manually change size and animation values

VitaliyBondarenko1982 (approved):
These tests sometimes failed even app work correctly
Looks  good.

PR #561, opened 2023-10-30
dev

nikkias:
[DEMO LINK](https://nikkias.github.io/react_carousel/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
       infinite={false}
     />
     ```
-    
+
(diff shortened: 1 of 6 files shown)


sTorba24 (changes requested):
Well done! Lets improve it
you need to recalculate transform after changing size
1 scroll then resized item
image

2. infinite mode should work in both sides (clicking prev too)

Andriy-Shymkiv (approved):
looks good

PR #564, opened 2023-11-02
add task solution

ydem1:
[DEMO LINK](https://ydem1.github.io/react_carousel/)


src/App.tsx
```diff
@@ -20,17 +29,192 @@ class App extends React.Component<{}, State> {
...
+            <select
+              name="itemsWidthSelect"
+              onChange={this.handleAnimationDurationChange}
+              value={selectedAnimationDuration}
+            >
+              {
+                animationsDuration.map((item, index) => (
+                  <option
+                    value={item}
+                    key={`${index + 1}`}
```
lerastarynets, line 201: don't use array indexes at all for key prop, consider something like [uuid](https://www.npmjs.com/package/uuid) instead
**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

lerastarynets (changes requested):
Pls disable right button, when u cannot scroll to right and same for left button if u cannot scroll to left

xiimiik (approved):
nice work!

PR #565, opened 2023-11-03
develop

atrofymovych:
[DEMO LINK](https://<your_account>.github.io/react_carousel/)


sTorba24 (changes requested):
Nice start!

These settings means that i should see only 3 elements and to be able to scroll them
image

but now all 10 images visible

PR #573, opened 2023-11-10
React Carousel

IvanOrinchak:
[DEMO LINK](https://IvanOrinchak.github.io/react_carousel/)


maxim2310 (changes requested):
Good job! Let's improve your code a bit)

1. it must be not placeholder but value 
image
2.  after one step i change size and something vent vrong
image

PR #584, opened 2023-11-22
Solution

92BS:
[DEMO LINK](https://92BS.github.io/react_carousel/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
       infinite={false}
     />
     ```
-    
+
 ## REQUIREMENTS:
 
 1. The title of the page should contain "Carousel"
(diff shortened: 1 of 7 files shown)


lerastarynets (changes requested):
Unfortunately, your project is not reachable via the demo link(

IvanFesenko (changes requested):
Good job. Check this issue:
https://github.com/mate-academy/react_carousel/assets/61904995/99da2924-4e51-49a9-87c9-86be69d4ebcd

92BS:
I just checked from my side and frame-size work, all is good, please check "Fix-Solution" maybe it will be a correct version

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
The right and left arrows are not working, when I click the right arrow, I don't get any image, just a white area
image

vadiimvooo (changes requested):
Good job, you are almost done with the task. Here is one interesting bug to fix:
1. Here are some bug with your carousel when you are scrolling to the end and adding frame size param:
https://www.loom.com/share/8d64f72f1b2b44078f08a2bb9771fffa

92BS:
@vadiimvooo, I just checked and and don't see this bug, add file 
Screenshot 2024-02-03 at 20 10 22

Andriy-Shymkiv (changes requested):
It is strange to be honest, because got the same bug as mentor above.
@92BS Can you please check this behavior on a different browser/computer?

TarasHoliuk (approved):
Well done :heavy_check_mark:

PR #585, opened 2023-11-22
I need help

SerhiiUnhurian:
[DEMO LINK](https://SerhiiUnhurian.github.io/react_carousel/)


denys-danyliuk (changes requested):
your demo link isn't working

SerhiiUnhurian:
Screenshot 2023-11-23 at 11 26 52

SerhiiUnhurian:
i cant make this deploy

SerhiiUnhurian:
check it out now

lerastarynets (changes requested):
Unfortunately the project via the demo link isn't working
Screenshot at Nov 23 12-17-24

SerhiiUnhurian:
> Unfortunately the project via the demo link isn't working Screenshot at Nov 23 12-17-24

yes i know
i need help with it

volodymyr-soltys97 (changes requested):
The item width doesn't work, feel free in the fe_chat if you need a help
image

src/App.scss
```diff
@@ -1 +1,12 @@
 // style go here
```
OleziO, line 1: Remove all comments from the code.

src/components/Carousel.scss
```diff
@@ -1 +1,57 @@
...
+
+  // &__wrapper {
+  //   width: 260px;
+  //   overflow: hidden;
+  // }
+
+  // &__wrapper {
+  //   // display: inline-block;
+  //   // transition: transform 0.3s ease-in-out;
+  // }
```
OleziO, line 18: Remove all comments from the code.

OleziO (changes requested):
image
Carousel and inputs do not work

src/components/Carousel.scss
SerhiiUnhurian: okay

src/App.tsx
```diff
@@ -20,17 +25,102 @@ class App extends React.Component<{}, State> {
...
         {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1 data-cy="title">Carousel with {images.length} images</h1>
+
+        <div>
+          <label htmlFor="itemWidth">Item Width: </label>
+          <input
+            type="number"
+            value={itemWidth}
+            onChange={(event) => this.handleInputChange(
```
anastasiiavorobiova, line 69:
```suggestion
            onChange={this.handleInputChange}
```

src/App.tsx
```diff
@@ -20,17 +25,102 @@ class App extends React.Component<{}, State> {
...
     return (
       <div className="App">
         {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1 data-cy="title">Carousel with {images.length} images</h1>
+
+        <div>
+          <label htmlFor="itemWidth">Item Width: </label>
+          <input
+            type="number"
```
anastasiiavorobiova, line 67:
```suggestion
            type="number"
            name="itemWidth"
```

src/App.tsx
```diff
@@ -20,17 +25,102 @@ class App extends React.Component<{}, State> {
...
+    event:
+    React.ChangeEvent<HTMLInputElement>, key: keyof State,
+  ) => {
+    const value = key === 'infinite'
+      ? event.target.checked
+      : Number(event.target.value);
+
+    this.setState((prevState) => ({
+      ...prevState,
+      [key]: Number.isNaN(value) ? 0 : value,
```
anastasiiavorobiova, line 45:
```suggestion
      [event.target.name]: Number.isNaN(value) ? 0 : value,
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/App.tsx
SerhiiUnhurian: but how will it work?
don't we need this event?

src/App.tsx
SerhiiUnhurian: i need to delete type?

src/App.tsx
```diff
@@ -20,17 +25,108 @@ class App extends React.Component<{}, State> {
...
+    React.ChangeEvent<HTMLInputElement>, key: keyof State,
+  ) => {
+    const value = key === 'infinite'
+      ? event.target.checked
+      : Number(event.target.value);
+
+    this.setState((prevState) => ({
+      ...prevState,
+      [event.target.name]: Number.isNaN(value) ? 0 : value,
+      // [key]: Number.isNaN(value) ? 0 : value,
```
volodymyr-soltys97, line 46: Remove comments
```suggestion
```

src/App.tsx
```diff
@@ -20,17 +25,108 @@ class App extends React.Component<{}, State> {
...
     return (
       <div className="App">
         {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1 data-cy="title">Carousel with {images.length} images</h1>
+
+        <div>
+          <label htmlFor="itemWidth">Item Width: </label>
+          <input
+            // type="number"
```
volodymyr-soltys97, line 68: The same as above
```suggestion
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,96 @@
...
+        ? newIndex
+        : images.length - frameSize);
+    }
+
+    setTransitionDuration(animationDuration);
+  };
+
+  return (
+    <div className="Carousel" style={{ width: widthContainer }}>
+      {/* <h1 data-cy="title">Carousel</h1> */}
```
volodymyr-soltys97, line 55:
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Try to place the buttons for switching on the same level
Also when I select infinite slider the toggle button doesn't activate, if you haven't implemented this logic then it's better to remove this checkbox
image

src/components/Carousel.tsx
SerhiiUnhurian: please help me to fix infinite scroll if we need it here

src/App.tsx
TarasHoliuk: at least now you also pass `itemWidth` as a separate argument, so it wouldn't work like this.

but code like this `onChange={(event) => this.handleInputChange(event)}` doesn't make sense it's equal to `onChange={this.handleInputChange}`. 

Explanation: to onChange prop you pass the function that accepts `event` as a prop. In the first variant, you pass an additional wrapper function, which inside calls the original function.

P.S. You may get `itemWidth` from `event.target.name`

src/App.tsx
```diff
@@ -20,17 +25,103 @@ class App extends React.Component<{}, State> {
...
+        </div>
 
-        <Carousel />
+        <Carousel
+          images={images}
+          step={step}
+          frameSize={frameSize}
+          itemWidth={itemWidth}
+          animationDuration={animationDuration}
+          infinite={false}
```
TarasHoliuk, line 123: there is still static `false` instead of dynamic variable
You haven't pushed/commited the updated code properly

TarasHoliuk (changes requested):
It seems you haven't pushed required changes (and haven't re-deployed solution) - please, do it and re-request review again

loralevitska (changes requested):
Almost done) Let's improve some bugs
- fix this bug while `infinite` checkbox checked 
image 
- add max and min values for each field to avoid such bugs
image
image

loralevitska (approved):
Almost perfect) 
`max` value still doesn't work, please fix it for improvement
image

PR #586, opened 2023-11-24
Carousel_OkM

Oksana-Maruniak:
[DEMO LINK](https://Oksana-Maruniak.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -30,6 +30,47 @@ class App extends React.Component<{}, State> {
         {/* eslint-disable-next-line */}
         <h1>Carousel with {images.length} images</h1>
 
+        <form className="App__form" action="/">
+
+          <label className="App__label">
+            Box Width:
+            <input
+              className="App__input"
+              type="number"
+              min={130}
+            />
+          </label>
+
+          <label className="App__label">
+            Frame Size:
+            <input
+              className="App__input"
+              type="number"
+              min={1}
+              max={10}
+            />
+          </label>
+
+          <label className="App__label">
+            Step:
+            <input
+              className="App__input"
+              type="number"
+              min={1}
+            />
+          </label>
+
+          <label className="App__label">
+            Animation Duration:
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,39 +1,143 @@
...
+  './img/9.png',
+  './img/10.png',
+];
+
+const App: React.FC = () => {
+  const [itemWidth, setBoxWidth] = useState(130);
+  const [frameSize, setFrameSize] = useState(3);
+  const [step, setStep] = useState(3);
+  const [animationDuration, setAnimationDuration] = useState(1000);
+  const [infinite, setInfinite] = useState(false);
```
VitaliyBondarenko1982, line 23: make one state and one handler for all inputs. use event.target.name for change proper field in state.
```suggestion
const [inputs, setInputs] = useState({
  itemWidth:  130,
  frameSize: 3,
  ...
})
```
value for name attibute in input should be the same as field in inputs object.

VitaliyBondarenko1982 (changes requested):
Looks good.
Please check comment.

src/App.tsx
Oksana-Maruniak: But why? We now have different processes running separately. Instead, we will get one object that changes completely for the one of inputs. Is not it?

src/App.tsx
```diff
@@ -1,39 +1,130 @@
...
+    event: React.ChangeEvent<HTMLInputElement>,
+  ) => {
+    const value = event.target.name === 'infinite'
+      ? event.target.checked
+      : event.target.value;
+
+    setInputs({
+      ...inputs,
+      [event.target.name]: value,
+    });
```
loralevitska, line 39:
```suggestion
    setInputs(prevState => ( {
      ...prevState,
      [event.target.name]: value,
    });
```

loralevitska (approved):
Good job!

PR #589, opened 2023-11-26
made solution for carousel

MrHunterBoi:
[DEMO LINK](https://MrHunterBoi.github.io/react_carousel/)

Also I legit don't know why some tests return with undefined even though others pass just fine

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,76 @@
-import React from 'react';
+import { useState } from 'react';
 import './App.scss';
 import Carousel from './components/Carousel';
 
-interface State {
-  images: string[];
-}
+function App() {
+  const images = [
+    './img/1.png',
+    './img/2.png',
+    './img/3.png',
+    './img/4.png',
+    './img/5.png',
+    './img/6.png',
+    './img/7.png',
+    './img/8.png',
+    './img/9.png',
+    './img/10.png',
+  ];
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,76 @@
...
+    './img/7.png',
+    './img/8.png',
+    './img/9.png',
+    './img/10.png',
+  ];
+
+  const [itemWidth, setItemWidth] = useState(130);
+  const [frameSize, setFrameSize] = useState(3);
+  const [step, setStep] = useState(3);
+  const [animationDuration, setAnimationDuration] = useState(1000);
```
ArtemTeslenko, line 22: use CONSTANTS for initial values

src/components/Carousel.tsx
```diff
@@ -1,18 +1,74 @@
...
+      <ul
+        className="Carousel__list"
+        style={{
+          transform: `translateX(${translation}px)`,
+          maxWidth: `${widthOfAllImages}px`,
+          transition: `transform ${animationDuration}ms`,
+        }}
+      >
+        {images.map((image, index) => (
+          <li key={`${index + 1}`}>
```
ArtemTeslenko, line 51: do not use index as keys, here you can use img src as keys, they are unique and stable

ArtemTeslenko (approved):
Good job 👍

PR #592, opened 2023-12-01
add Solution

Softjey:
[DEMO LINK](https://softjey.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel/Carousel.tsx
+++ b/src/components/Carousel/Carousel.tsx
@@ -18,48 +18,46 @@ const Carousel: React.FC<Props> = ({
   animationDuration,
   infinite,
 }) => {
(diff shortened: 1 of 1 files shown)


sTorba24 (changes requested):
Well done! Lets improve it

Looks broken when i change frame size after scroll
image

sTorba24 (changes requested):
Good work!
it means that i should have 5 visible items with size 200px, but in result it's wrong
image

maxim2310 (changes requested):

1. if you reach the last img and start to increase frame size you will get this 
image
2. and when you change frame size you force carousel back on 1 step, thats strange

maxim2310 (approved):
Well done!!!

PR #594, opened 2023-12-02
Solution

Hladkyi03:
[DEMO LINK](https://Hladkyi03.github.io/react_carousel/)
Idk, why test are falling, they aren't run at all. But you can see the functionality on demo link :)
I also added navbar, that will works, only if step = frameSize, in other cases when they aren't similar i don't see the point to implement navbar

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,10 +4,15 @@ import Carousel from './components/Carousel';
 
 interface State {
   images: string[];
+  step: number;
+  frameSize: number;
+  itemWidth: number;
+  animationDuration: number;
+  infinite: boolean;
 }
 
 class App extends React.Component<{}, State> {
-  state = {
+  state: State = {
     images: [
       './img/1.png',
       './img/2.png',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -20,17 +25,133 @@ class App extends React.Component<{}, State> {
...
     return (
       <div className="App">
-        {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1 data-cy="title" className="title">
+          Carousel with
+          {' '}
+          {images.length}
+          {' '}
+          images
```
VitaliyBondarenko1982, line 71:
```suggestion
          {`Carousel with ${images.length} images`}
```

src/App.tsx
```diff
@@ -20,17 +25,133 @@ class App extends React.Component<{}, State> {
...
+            <label htmlFor="itemWidth" className="settings__label">
+              Item width:
+            </label>
+            <input
+              type="number"
+              className="settings__input"
+              name="itemWidth"
+              step={10}
+              value={itemWidth}
+              onChange={(e) => this.handleInputChange(e, 'itemWidth')}
```
VitaliyBondarenko1982, line 87:
```suggestion
              onChange={this.handleInputChange}
```
you can use event.target.name which will be equal value of name attribute for change proper field in state

src/App.tsx
```diff
@@ -20,17 +25,133 @@ class App extends React.Component<{}, State> {
...
+    e: React.ChangeEvent<HTMLInputElement>,
+    key: keyof State,
+  ) => {
+    const { value } = e.target;
+
+    this.setState(prevState => ({
+      ...prevState,
+      [key]: +value,
+    }));
+  };
```
VitaliyBondarenko1982, line 45: you can make one handler for all inputs.
something like that.

```suggestion
  handleInputChange = (
    e: React.ChangeEvent<HTMLInputElement>,
  ) => {
    const { value, name, checked } = e.target;

    this.setState(prevState => ({
      ...prevState,
      [name]: name === 'infinite' ? checked : +value,
    }));
  };
```

VitaliyBondarenko1982 (changes requested):
Good job.
Let's try to fix comments and such behavior (see screens).
Then will see problems with tests
- min value for inputs should be limited 
chrome-capture-2023-11-2 (1)
Screenshot from 2023-12-02 18-21-59
If you will have some questions - ask in chat.

Hladkyi03:
Tests still doesn't running, but now functionality implemented correctly

volodymyr-soltys97 (approved):
Good job 👍

PR #597, opened 2023-12-04
Solution

patrykyo20:
https://Patrykmclaren14.github.io/react_carousel/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,6 +24,8 @@ class App extends React.Component<{}, CarouselProps> {
     infinite: false,
   };
 
+  debounceTimeout: NodeJS.Timeout | null = null;
+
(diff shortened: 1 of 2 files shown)


Barosz30 (approved):
Could be styled a bit, but on technical side, it's great

mbruhler (changes requested):
In overall, it works good on technical side but in my opinion there should be some tweaks introduced:

When I delete all numbers from input, it forces 0
image
Then I can't delete it and providing new number results in '0' prefixed numbers

EXTRA (not needed to be approved, but nice to have):
Also maybe add a little debounce for inputs to prevent jumping whole UI when filling the form

mvjl000 (approved):
Well done!

PR #601, opened 2023-12-08
Task solution

Shostyy:
[DEMO LINK](https://shostyy.github.io/react_carousel/)


src/App.tsx
```diff
@@ -20,17 +25,137 @@ class App extends React.Component<{}, State> {
...
     ],
+    itemWidth: 130,
+    frameSize: 3,
+    step: 3,
+    animationDuration: 1000,
+    infinite: false,
+  };
+
+  handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    switch (event.target.id) {
```
TarasHoliuk, line 36: just FYI - you can also add `name` attribute to your `inputs` instead of id, and read it here
often you would anyway have `name` attribute

TarasHoliuk (approved):
GJ - Approved :heavy_check_mark:

TarasHoliuk (changes requested):
GJ :+1: Let's improve some things - style change handling

> I skipped an optional task for infinity scroll option, and from what I saw - no one so far has done it properly.

sure, you may not do it now, but just FYI (it seems it often is misunderstood): the correct way to implement "infinite" in this task is the following: if the flag is enabled and you are in the start of the carousel the `prev` button shouldn't be disabled - it should scroll to the end of carousel. The same with the `next` button. It's a simplified version of infinite scroll. In the PR before your PR it was implemented in the correct way - as far as I understood you've checked a few demos

Shostyy:
I'm sory, i realize it only right now. This comment looks kinda like it's was written for me, but i think it's not. So idk why i fixid my solution. I'll do reset)

VitaliyBondarenko1982 (approved):

PR #604, opened 2023-12-13
Solution

ghost:
[DEMO LINK](https://OstTomash.github.io/react_carousel/)


maxim2310 (changes requested):
Good work, but need to check frame, because now if i click next and try to increase a size of img it looks like 
image

PR #605, opened 2023-12-14
v1

DVasylenko18:
[DEMO LINK](https://DVasylenko18.github.io/react_carousel/)


maxim2310 (changes requested):
Good work, lets make some improvement 
1. the main problem that carousel must move smooth without jumping on click
2. when i increase steps it move img that shouldn't be 
3. when i increase img size it must increase width and height proportional but it isn't

lerastarynets (approved):
You've got this, keep pushing forward!

PR #610, opened 2023-12-20
solution

AnastasiiaKriuchkova:
[DEMO LINK](https://AnastasiiaKriuchkova.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,106 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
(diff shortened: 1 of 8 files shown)


sTorba24 (changes requested):
Good work! Lets improve it, check your infinite mode, it's not working :(
image

src/components/Carousel.tsx
```diff
@@ -1,18 +1,104 @@
...
+  infinite,
+}) => {
+  const [index, setIndex] = useState(0);
+
+  const maxIndex = images.length - frameSize;
+  const atStart = index === 0 && !infinite;
+  const atEnd = index === images.length - frameSize && !infinite;
+
+  const handleNext = () => {
+    if (index < maxIndex) {
```
TarasHoliuk, line 28: you should take into account the `infinity` property here:

if now `index === 0` and `infinite === true` you should scroll to the end

The similar with `handlePrev` function

TarasHoliuk (changes requested):
Hey! Indeed, the code looks good, but `infinity` prop is still not handled as expected

If something is unclear feel free to ask in the chat :)

sTorba24 (approved):
Looks cool now!

PR #613, opened 2023-12-21
solution

GavriliukArtem:
[DEMO LINK](https://GavriliukArtem.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,83 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li><img src="./img/1.png" alt="1" /></li>
-      <li><img src="./img/1.png" alt="2" /></li>
-      <li><img src="./img/1.png" alt="3" /></li>
-      <li><img src="./img/1.png" alt="4" /></li>
-    </ul>
+type Props = {
+  images: string[];
+  itemWidth: number;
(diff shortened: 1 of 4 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,83 @@
...
+    transform: `translateX(${-(imgPosition * itemWidth)}px)`,
+    transition: `transform ${animationDuration}ms ease`,
+  };
+
+  const handlePrev = () => {
+    if (imgPosition > 0) {
+      setImgPosition((prev) => (prev - step >= 0 ? prev - step : 0));
+    } else {
+      setImgPosition(lastImgPosition);
+    }
```
Viktor-Kost, line 41: you can use ternary inside `setImgPosition` to avoid using `if`/`else`

src/components/Carousel.tsx
```diff
@@ -1,18 +1,83 @@
...
+  };
+
+  const handleNext = () => {
+    if (imgPosition < lastImgPosition) {
+      setImgPosition((prev) => (prev + step <= lastImgPosition
+        ? prev + step
+        : lastImgPosition
+      ));
+    } else {
+      setImgPosition(0);
```
Viktor-Kost, line 51: it's better to use early return instead `if/else`
```suggestion
    if (imgPosition < lastImgPosition) {
      setImgPosition((prev) => (prev + step <= lastImgPosition
        ? prev + step
        : lastImgPosition
      ));
      
      return;
    }
    
    setImgPosition(0);
```

Viktor-Kost (approved):
GJ! 🔥

PR #620, opened 2023-12-30
solution

RomanO3GIT:
some tests are failing, can you please check where the is issue? Thank you!

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,18 +1,100 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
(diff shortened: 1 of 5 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,100 @@
...
+  const carouselStyle = {
+    width: `${itemWidth * frameSize}px`,
+  };
+
+  const carouselImageStyle = {
+    transform: `translateX(${-(imgPosition * itemWidth)}px)`,
+    transition: `transform ${animationDuration}ms ease`,
+  };
+
+  const handlePrev = () => {
```
TarasHoliuk, line 36: does it work with `infinite` properly?

For info, if `infinite` is enabled and the current position is 0 you should be able to click the button and images should move to the last position

The similar for `next` button

TarasHoliuk (changes requested):
Please, fix the DEMO LINK (404 page now)

PR #622, opened 2024-01-04
add finished project

allagutsul27:
[DEMO LINK](https://allagutsul27.github.io/react_carousel/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,27 @@
-// style go here
+.App {
+  &__title {
+    text-align: center;
+    font-family: Verdana, Geneva, Tahoma, sans-serif;
+    font-style: italic;
+  }
+
+  &__labels {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    font-size: 16px;
+    gap: 20px;
+    & input {
+      margin-left: 10px;
+      text-align: center;
+      font-weight: bold;
+      padding: 5px;
+      background-color: rgb(234, 255, 191);
+      color: rgb(69, 75, 75);
+      border: 1px solid black;
+    }
+    & .infinite {
+      display: flex;
+    }
+  }
+}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,89 @@
-import React from 'react';
+import { useState } from 'react';
 import './App.scss';
 import Carousel from './components/Carousel';
+import { images } from './images';
 
-interface State {
-  images: string[];
-}
+export const App = () => {
+  const [step, setStep] = useState(3);
+  const [frameSize, setFrameSize] = useState(3);
+  const [itemWidth, setItemWidth] = useState(130);
+  const [animationDuration, setAnimationDuration] = useState(1000);
+  const [infinite, setInfinite] = useState(false);
 
-class App extends React.Component<{}, State> {
-  state = {
-    images: [
-      './img/1.png',
-      './img/2.png',
-      './img/3.png',
-      './img/4.png',
-      './img/5.png',
-      './img/6.png',
-      './img/7.png',
-      './img/8.png',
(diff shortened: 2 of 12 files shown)


maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

1. would be nice to add some style for alignment(optional)  
image
2. in the end i increase frame size and ...
image

src/App.tsx
```diff
@@ -1,39 +1,89 @@
...
-        {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
-
-        <Carousel />
+export const App = () => {
+  const [step, setStep] = useState(3);
+  const [frameSize, setFrameSize] = useState(3);
+  const [itemWidth, setItemWidth] = useState(130);
+  const [animationDuration, setAnimationDuration] = useState(1000);
+  const [infinite, setInfinite] = useState(false);
```
Viktor-Kost, line 11: it's much better to make it as one instance similar you did with movie add form

1. Create a variable outside the Component
```
const initialOptionsState = {
  step: 3,
  frameSize: 3,
  itemWidth: 130,
  animationDuration: 1000,
  infinite: false,
}
```

2. Use one state for options
3. Whenever you need to change value of any field you should pass two arguments `(key, value)` and then just change a value of a movie object by key. In this case you will reuse one handler instead of handlers for each input
Example: 

```
const handleOptionChange = (key, value) => {
  setOptions(prevOptions => ({...prevOptions, [key]: value}))
}
```

```suggestion
  const [options, setOptions] = useState(initialOptionsState);
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,96 @@
...
+  useEffect(() => {
+    setPosition(0);
+  }, [frameSize]);
+
+  const handleNextClick = () => {
+    if (position + step < lastPosition) {
+      setPosition(position + step);
+    } else {
+      setPosition(lastPosition);
+    }
```
Viktor-Kost, line 30: you can use ternary inside 
Like:
```
setPosition(condition ? value1 : value2)
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,96 @@
...
+      setPosition(0);
+    }
+  };
+
+  const handlePrevClick = () => {
+    if (position - step < 0) {
+      setPosition(0);
+    } else {
+      setPosition(position - step);
+    }
```
Viktor-Kost, line 42: the same is about ternary

src/App.scss
```diff
@@ -1 +1,27 @@
-// style go here
+.App {
+  &__title {
+    text-align: center;
+    font-family: Verdana, Geneva, Tahoma, sans-serif;
+    font-style: italic;
+  }
+
+  &__labels {
```
Viktor-Kost, line 8: That what previous mentor said (optionally)
Screenshot 2024-01-05 at 16 50 50

Viktor-Kost (changes requested):
Great job!
Let's improve your code since solution works well

src/App.tsx
```diff
@@ -1,39 +1,109 @@
...
+        </label>
+
+        <label>
+          Frame size:
+          <input
+            type="number"
+            min={1}
+            max={10}
+            value={frameSize}
+            onChange={(e) => handleOptionChange(
```
loralevitska, line 53: It can be improved:
```suggestion
            onChange={handleOptionChange}
```
Set names to input and use one handler

```
const handleOptionChange = (e) => {
    const { name, value } = e.target
    setOptions(prevOptions => ({ ...prevOptions, [name]: value }));
  };
  ```

loralevitska (approved):
Good job!

PR #630, opened 2024-01-16
first commit

FallenMAD:
[DEMO LINK](https://FallenMAD.github.io/react_carousel/)


maxim2310 (changes requested):
1. i scroll till end, then increase frame size 
image
2. need to set max value for frame size and check other inputs 
image
3. need to add steps 10px for item size

PR #635, opened 2024-01-23
Add task solution

PodvAx:
[DEMO LINK](https://PodvAx.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,160 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import Carousel from './components/Carousel';
+import {
+  moveCarouselItemToPosition,
+  MAX_ANIMATION_DURATION,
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,39 +1,160 @@
...
+            id="step"
+            className="input-section__input"
+            value={step}
+            min="1"
+            max={frameSize}
+            onChange={(event) => {
+              moveCarouselItemToPosition(0);
+              setStep(Number(event.target.value));
+            }}
+          />
```
lerastarynets, line 78: Pls move the function inside `onChange` handler to a separate const and use as a callback here
**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #641, opened 2024-01-29
Solution

vasylynaBC:
[DEMO LINK](https://vasylynaBC.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,108 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import Carousel from './components/Carousel';
 
-interface State {
-  images: string[];
-}
+const images = [
+  './img/1.png',
+  './img/2.png',
+  './img/3.png',
+  './img/4.png',
+  './img/5.png',
+  './img/6.png',
+  './img/7.png',
+  './img/8.png',
+  './img/9.png',
+  './img/10.png',
+];
 
-class App extends React.Component<{}, State> {
-  state = {
-    images: [
-      './img/1.png',
-      './img/2.png',
-      './img/3.png',
-      './img/4.png',
-      './img/5.png',
-      './img/6.png',
-      './img/7.png',
-      './img/8.png',
-      './img/9.png',
-      './img/10.png',
-    ],
-  };
+const App: React.FC = () => {
+  const [itemWidth, setItemWidth] = useState(130);
+  const [size, setSize] = useState(3);
+  const [step, setStep] = useState(3);
+  const [duration, setDuration] = useState(1000);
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,39 +1,108 @@
...
-      './img/9.png',
-      './img/10.png',
-    ],
-  };
+const App: React.FC = () => {
+  const [itemWidth, setItemWidth] = useState(130);
+  const [size, setSize] = useState(3);
+  const [step, setStep] = useState(3);
+  const [duration, setDuration] = useState(1000);
+  const [infinite, setInfinite] = useState(false);
```
Viktor-Kost, line 23: It's not advisable to generate such a large number of states when we only have one instance (settings). 
It'd be much better to utilize a single state with an object inside.

1. Create a variable outside the Component
```
const initialSettingsState = {
  itemWidth: 130,
  size: 3,
  step: 3,
  duration: 1000,
  infinite: false
}
```

2. Use one state for settings
3. Whenever you need to change value of any field you should pass two arguments `(key, value)` and then just change a value of a settings object by key. In this case you will reuse one handler instead of handlers for each input
Example: 

```
const handleInputChange = (key, value) => {
  setSettings(prevInputs => ({...prevSettings, [key]: value}))
}
```


```suggestion
  const [settings, setSettings] = useState(initialSettingsState);
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,102 @@
...
+  const maxImages = images.length - frameSize;
+  const disableAction = !infinite && currentImage === 0;
+  const disableActionNext = !infinite && currentImage === maxImages;
+
+  const prevItems = () => {
+    if (currentImage === 0) {
+      setCurrentImage(maxImages);
+    } else {
+      setCurrentImage(prev => Math.max((prev - step), 0));
+    }
```
Viktor-Kost, line 32: use ternary operator inside setCurrentImage instead

src/components/Carousel.tsx
```diff
@@ -1,18 +1,102 @@
...
+  };
+
+  const nextItems = () => {
+    if (currentImage !== maxImages) {
+      setCurrentImage(prev => Math.min(prev + step, maxImages));
+    } else {
+      setCurrentImage(0);
+    }
```
Viktor-Kost, line 40: ternary

Viktor-Kost (changes requested):
Also please fix this issue

https://github.com/mate-academy/react_carousel/assets/117587463/4e5a07ec-166c-4a66-a465-2a627a907eec

lerastarynets (approved):
lgtm 🚀

PR #643, opened 2024-01-30
Solution to task

tniklevych:
- [DEMO LINK](https://tniklevych.github.io/react_carousel/)


maxim2310 (changes requested):

1. would be nice to align it a bit 
image
2. i scrolled till end and increase frame size 
image

src/components/Carousel/Carousel.scss
```diff
@@ -0,0 +1,60 @@
...
+  &__btn {
+    margin: 0 10px;
+    width: 60px;
+    height: 30px;
+
+    background: none;
+    border: 2px solid #178acb;
+    border-radius: 5px;
+    color: #fff;
+    -webkit-box-shadow:inset 0 0 0 100px #178acb;
```
OleziO, line 31: Move all repeatable colors to the CSS variables.

[More info](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

OleziO (changes requested):

src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,125 @@
...
+              />
+            </li>
+          ))}
+        </ul>
+      </div>
+
+      <div className="Carousel__btns">
+        <button
+          type="button"
+          className={`Carousel__btn ${isDisabledButtonPrev ? 'Carousel__btn-disabled' : ''}`}
```
volodymyr-soltys97, line 105: Need to use the classnames library for add classes with condition

volodymyr-soltys97 (approved):
Good job 👍

PR #644, opened 2024-01-30
carouse 1.0

EvgenGolub:


src/App.tsx
```diff
@@ -1,39 +1,130 @@
...
+  './img/4.png',
+  './img/5.png',
+  './img/6.png',
+  './img/7.png',
+  './img/8.png',
+  './img/9.png',
+  './img/10.png',
+]
+
+enum DefaultValues {
```
TarasHoliuk, line 18: if you want to store **values** it's better to use object
BTW, such constants are often named in UPPER case:

`const DEFAULT_VALUES = {}`

src/components/Carousel.tsx
```diff
@@ -1,18 +1,92 @@
...
+  return (
+    <div className="Carousel">
+      <ul className="Carousel__list">
+        {images.map((image, i) => (
+          <li
+            className='Carousel__image'
+          >
+            <img
+              src={image}
+              alt={`${i + 1}`}
```
TarasHoliuk, line 55: just index is not a good `alt` attribute content
I understand that there is not a lot of info) But you may write smth like that: `Emoji image #${i + 1}`, etc

src/components/Carousel.tsx
```diff
@@ -1,18 +1,92 @@
...
+              width={itemWidth}
+            />
+            </li>
+        ))}
+      </ul>
+      <div className="Carousel__buttons">
+        <button
+          type="button"
+          className={
```
TarasHoliuk, line 64: not critical, but it's better to use classnames for such cases

TarasHoliuk (changes requested):
The deployed solution doesn't work :cry: 
image

Fix it to make review possible (ask in the chat if you need help)

Viktor-Kost (changes requested):
Please check your solution since settings (besides first one) don't work

https://github.com/mate-academy/react_carousel/assets/117587463/fe25fabe-86a7-49c5-b54b-e9226a49b786

EvgenGolub:
[DEMO LINK](https://evgengolub.github.io/react_carousel/)

src/App.tsx
```diff
@@ -1,39 +1,135 @@
...
+  const [infinite, setInfinite] = useState(false);
+  const [currentImage, setCurrentImage] = useState(0);
+
+  return (
+    <div className="App">
+      <h1 data-cy="title" className="App__title">
+        Carousel with
+        {images.length}
+        {' '}
+        images
```
vadiimvooo, line 39: You can simplify this lines to such one:

```suggestion
       {`Carousel with ${images.length} ${images}`}
```

src/components/Carousel.tsx
```diff
@@ -1,18 +1,105 @@
...
+        ? setCurrentImage(currentImage - step >= 0 ? currentImage - step : 0)
+        : setCurrentImage(imagesLeft)
+    );
+  }
+
+  function goNext() {
+    return (
+      currentImage !== imagesLeft
+        ? setCurrentImage(
+          currentImage + step <= imagesLeft ? currentImage + step : imagesLeft,
```
vadiimvooo, line 39:
```suggestion
          currentImage + step <= imagesLeft 
                ? currentImage + step 
                : imagesLeft,
```

src/components/Carousel.tsx
vadiimvooo: I agree with the previous comment.

src/App.tsx
```diff
@@ -1,39 +1,135 @@
...
+  './img/8.png',
+  './img/9.png',
+  './img/10.png',
+];
+
+const DEFAULT_VALUES = {
+  itemWidth: 130,
+  frameSize: 3,
+  step: 3,
+  animDur: 1000,
```
vadiimvooo, line 22: it's better to use here full name of variable for better understanding
```suggestion
  animationDuration: 1000,
```

vadiimvooo (changes requested):
In general good job, everything looks fine just a few changes required,

TarasHoliuk (approved):

PR #648, opened 2024-02-08
solution

Dmytro-Kretov:
[DEMO LINK](https://dmytro-kretov.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -28,9 +28,18 @@ class App extends React.Component<{}, State> {
     return (
       <div className="App">
         {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <div className="App__container">
+          <h1
+            data-cy="title"
+            className="title"
+          >
+            {`Carousel with ${images.length} images`}
+          </h1>
 
-        <Carousel />
+          <Carousel
+            images={images}
+          />
+        </div>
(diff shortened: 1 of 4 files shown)


src/components/Carousel.tsx
```diff
@@ -1,18 +1,171 @@
...
+          type="button"
+          disabled={xPos === maxItemsWidth + selectedWidth * selectedFrames}
+          className="button"
+          onClick={nextBTN}
+        >
+          Next
+        </button>
+      </div>
+
+      <div className="input">
```
maxim2310, line 98: would be nice to move form to separate component

src/App.tsx
```diff
@@ -28,9 +28,18 @@ class App extends React.Component<{}, State> {
...
+        <div className="App__container">
+          <h1
+            data-cy="title"
+            className="title"
+          >
+            {`Carousel with ${images.length} images`}
+          </h1>
 
-        <Carousel />
+          <Carousel
```
maxim2310, line 39: this component must be used like 
image

maxim2310 (changes requested):

1. looks not so good 
image
2.when i change item size position of imgs was reset, that not must be 
3. same with steps and frame size

Andriy-Shymkiv (approved):
good job

PR #656, opened 2024-02-16
develop

RomanSylman:
[DEMO LINK](https://RomanSylman.github.io/react_carousel/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,21 @@
-// style go here
+.app {
+
+  &__fields {
+    width: 500px;
+    display: flex;
+    flex-direction: column;
+    gap: 15px;
+
+    padding: 10px;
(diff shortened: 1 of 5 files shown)


maxim2310 (changes requested):
Almost there) Just couple more things needs to be fixed

1. when i change frame size, animation duration or item size carusel is reseted
2. i cant increase steps
3. would be nice to center carousel

RomanSylman:
> Almost there) Just couple more things needs to be fixed
> 
> 1. when i change frame size, animation duration or item size carusel is reseted
> 2. i cant increase steps
> 3. would be nice to center carousel

Ok, i'm just tested my DEMO link. And I don't find first two issues, next push will be centered.

SanyaBratashchuk (approved):

PR #660, opened 2024-02-17
Develop

ivankovbohdan:
[DEMO LINK](https://ivankovbohdan.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,109 @@
-import React from 'react';
-import './App.scss';
-import Carousel from './components/Carousel';
+import React, { useState } from "react";
+import "./App.scss";
+import Carousel from "./components/Carousel";
 
-interface State {
-  images: string[];
-}
+const App: React.FC = () => {
+  const images = [
+    "./img/1.png",
+    "./img/2.png",
+    "./img/3.png",
+    "./img/4.png",
+    "./img/5.png",
+    "./img/6.png",
+    "./img/7.png",
+    "./img/8.png",
+    "./img/9.png",
+    "./img/10.png",
+  ];
 
-class App extends React.Component<{}, State> {
-  state = {
-    images: [
-      './img/1.png',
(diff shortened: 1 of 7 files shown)


SanyaBratashchuk (changes requested):
In infinity mode when frame size is less than 3 I can scroll to empty space 
image

Viktor-Kost (changes requested):
We should see default setting values (since they're set)  
Screenshot 2024-02-19 at 14 45 03

src/App.tsx
```diff
@@ -1,39 +1,124 @@
...
+    frameSize: 3,
+    step: 3,
+    animationDuration: 1000,
+    infinite: false,
   };
 
-  render() {
-    const { images } = this.state;
+  const [settings, setSettings] = useState(initialSettingsState);
+  const [infinite, setInfinite] = useState(false);
```
Viktor-Kost, line 28: we've used infinite inside `settings`, haven't we?

Viktor-Kost (changes requested):
Great improvements!

Yet we have some troubles with settings
We have 10 cards, but it doesn't work with amount. Everything works fine when we have default setup, but when we change some of settings it becomes broken

Probably the reason is how you updated `prev`/`next` methods

If you need any help feel free to ask at fe_chat

https://github.com/mate-academy/react_carousel/assets/117587463/da5d2435-e12a-49e8-8c04-bbb837480d35

ivankovbohdan:
> Great improvements!
> 
> Yet we have some troubles with settings We have 10 cards, but it doesn't work with amount. Everything works fine when we have default setup, but when we change some of settings it becomes broken
> 
> Probably the reason is how you've updated `prev`/`next` methods
> 
> If you need any help feel free to ask at fe_chat
> 
>  Screen.Recording.2024-02-19.at.18.27.34.mov

SanyaBratashchuk (approved):

PR #663, opened 2024-02-18
Solution

wsadocha:
[DEMO LINK](https://wsadocha.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,14 @@
 import React from 'react';
 import './App.scss';
-import Carousel from './components/Carousel';
+import { Carousel } from './components/Carousel';
 
-interface State {
+export interface State {
   images: string[];
+  itemWidth: number;
+  frameSize: number;
+  step: number;
+  animationDuration: number;
+  infinite: boolean;
 }
 
 class App extends React.Component<{}, State> {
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -20,17 +25,106 @@ class App extends React.Component<{}, State> {
...
-        <Carousel />
+        <form className="App__form">
+          <label htmlFor="itemWidth" className="App__label">
+            Item width
+            <input
+              className="App__input"
+              type="number"
+              id="itemWidth"
+              defaultValue={itemWidth}
+              onChange={event => this.setState({
```
yevhenii-pyl, line 57: Let's create a separate handler that can accept the `key` from state that should be changed as an argument.

src/App.tsx
```diff
@@ -1,9 +1,14 @@
...
+export interface State {
   images: string[];
+  itemWidth: number;
+  frameSize: number;
+  step: number;
+  animationDuration: number;
+  infinite: boolean;
 }
 
 class App extends React.Component<{}, State> {
```
yevhenii-pyl, line 14: Let's stick to one conception: it's either class based components or functional components. 
Here you have a class component, and your carousel is a function. Could you change one of them?

src/App.tsx
```diff
@@ -20,17 +25,106 @@ class App extends React.Component<{}, State> {
...
+              })}
+            />
+          </label>
+
+          <label htmlFor="frameSize" className="App__label">
+            Frame size
+            <input
+              className="App__input"
+              type="number"
+              min={1}
```
yevhenii-pyl, line 68: Let's avoid having that `magic numbers` in the code. Could you create a variables for min and max? 
`const MAX_IMAGES = 10`, etc

yevhenii-pyl (changes requested):
Good job! 👍 
Let's just fix a few things and we are good to go!

yevhenii-pyl (approved):
Nailed it 🚀 

(As a quick advice: that constants you created for min || max || etc values should be declared outside the component. 
Here, we assume that they will be a part of env variables at .env file and later taken from process.env. So to make them closer to real-life example, let's just declare them outside. Definitely not a mistake here, just wonted to notice. 
Also, I assume that images should be a part of your state, as it was previously in a class based component. )

PR #666, opened 2024-02-22
Develop

mrsvolodya:
https://mrsvolodya.github.io/react_carousel/


etojeDenys (approved):
also better to add some min and max limits to inputs

PR #677, opened 2024-03-10
Solution

RUBIC0N5545:
https://RUBIC0N5545.github.io/react_carousel/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,7 @@
-import React from 'react';
+import React, { ChangeEvent } from 'react';
 import './App.scss';
 import Carousel from './components/Carousel';
-
-interface State {
-  images: string[];
-}
+import { State } from './state';
 
 class App extends React.Component<{}, State> {
   state = {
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -20,17 +17,114 @@ class App extends React.Component<{}, State> {
...
   render() {
     const { images } = this.state;
+    const { itemWidth } = this.state;
+    const { frameSize } = this.state;
+    const { step } = this.state;
+    const { animationDuration } = this.state;
+    const { infinite } = this.state;
+
+    const changeStep = (e: ChangeEvent<HTMLInputElement>) =>
+      this.setState({ step: +e.target.value });
```
TarasHoliuk, line 36:
```suggestion
    const handleChange = (e: ChangeEvent<HTMLInputElement>) =>
      this.setState({ [e.target.name]: +e.target.value });
```
Let's make 1 shared `handleChange` function (at least for number inputs, but you can also handle boolean too - check input type or name). Above is an example
You can add `name` attribute to `input` element and use it here. Note that some additional checks might be necessary to make TypeScript happy :)

TarasHoliuk (changes requested):
GJ :+1: 

Let's add 1 small improvement

vadiimvooo (approved):
LGTM

PR #679, opened 2024-03-12
solution

ValeriiaFomenko:
[DEMO LINK](https://valeriiafomenko.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,8 @@
 import React from 'react';
 import './App.scss';
-import Carousel from './components/Carousel';
+import Carousel, { CarouselProps } from './components/Carousel';
 
-interface State {
-  images: string[];
-}
-
-class App extends React.Component<{}, State> {
+class App extends React.Component<{}, CarouselProps> {
   state = {
     images: [
       './img/1.png',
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -20,17 +16,96 @@ class App extends React.Component<{}, State> {
...
+    step: 3,
+    animationDuration: 1000,
+    infinite: false,
   };
 
   render() {
-    const { images } = this.state;
+    // eslint-disable-next-line operator-linebreak
+    const { images, itemWidth, frameSize, step, animationDuration, infinite } =
+      this.state;
```
DarkMistyRoom, line 29:
```suggestion
    const { 
      images, 
      itemWidth, 
      frameSize, 
      step, 
      animationDuration, 
      infinite 
    } = this.state;
```

src/components/Carousel.tsx
```diff
@@ -1,26 +1,92 @@
...
+  animationDuration: number;
+  infinite: boolean;
+}
+
+const Carousel: React.FC<CarouselProps> = ({
+  images,
+  itemWidth = 130,
+  frameSize = 3,
+  step = 3,
+  animationDuration = 1000,
```
DarkMistyRoom, line 19: would be great to move all these "magic" number to constants

DarkMistyRoom (changes requested):
Good job! But it seems like Infinite mode works only for the left direction
image

src/App.tsx
etojeDenys: image

etojeDenys (changes requested):
fix this free space when I change "Frame size"

src/components/Carousel.tsx
```diff
@@ -1,26 +1,103 @@
...
+  frameSize: number;
+  step: number;
+  animationDuration: number;
+  infinite: boolean;
+}
+
+const DEFAULT_ITEM_WIDTH = 130;
+const DEFAULT_FRAME_SIZE = 3;
+const DEFAULT_STEP = 3;
+const DEFAULT_ANIMATION_DURATION = 1000;
```
DarkMistyRoom, line 17: would be great to move constants to a separate file

src/components/Carousel.tsx
```diff
@@ -1,26 +1,103 @@
...
+        <ul
+          className="Carousel__list"
+          style={{
+            transform: `translateX(-${translateX}px)`,
+            transition: `transform ${animationDuration}ms ease`,
+          }}
+        >
+          {images.map((image, index) => (
+            <li key={image} className="Carousel__item">
+              <img src={image} alt={`${index + 1}`} width={itemWidth} />
```
DarkMistyRoom, line 73:
```suggestion
              <img 
                src={image} 
                alt={`${index + 1}`} 
                width={itemWidth}
              />
```

DarkMistyRoom (approved):
Good job!

PR #680, opened 2024-03-12
solution

katya-sn:
https://katya-sn.github.io/react_carousel/

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,26 @@
-// style go here
+.App {
+  text-align: center;
+}
+
+.customize {
+  align-items: center;
+  width: 300px;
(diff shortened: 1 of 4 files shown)


maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

broken when i increased frame size 
image

IvanFesenko (changes requested):
Not fixed
https://github.com/mate-academy/react_carousel/assets/61904995/241b64b4-fa57-4335-bb57-c0a5d8b0d560

katya-sn:
I'm confused cause it works correctly  at localhost

PR #690, opened 2024-03-15
task solution

Devery11:
[DEMO LINK](https://devery11.github.io/react_carousel/)


l4st1m0za (changes requested):
I have a starnge bug, that if I will change the number of icons scrolled, from 3 to any other number, it will start to lag, and scroll much more, that I wanted. Reduce this bug

GUSILLUS (changes requested):
GJ! Try to fix linter errors and test problems
image


Just try to write a solution that will good with tests

volodymyr-soltys97 (changes requested):
Almost done!
The step cannot be less than 0
image

VitaliyBondarenko1982 (approved):
Good job

PR #691, opened 2024-03-15
add task solution

TetianaBukhenko:
[DEMO LINK](https://TetianaBukhenko.github.io/react_carousel/) 
Screenshot_2
Hello, firstly I must admit that I didn`t pass 4 tests, however, everything works as it should. I would be grateful for any feedback.


GUSILLUS (changes requested):
Hi! Better to ask this on a fe_chat or a Q&A. I think we can find a solution if you write this question there

vadiimvooo (approved):
Well done!)

PR #697, opened 2024-03-20
Develop

BoGoBooM:
[DEMO LINK](https://bogoboom.github.io/react_carousel/)


maxim2310 (changes requested):
looks like broken 
image

src/App.tsx
```diff
@@ -20,17 +24,94 @@ class App extends React.Component<{}, State> {
...
+      step: +event.target.value,
+    }));
+  };
+
+  handleAnimationDuration = (event: React.ChangeEvent<HTMLInputElement>) => {
+    this.setState(prevState => ({
+      ...prevState,
+      animationDuration: +event.target.value,
+    }));
   };
```
Viktor-Kost, line 59: You can simplify this handler and reuse for each input by adding `name`s to your inputs (or you can take `id`)
```suggestion
  handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
    const { id, value } = e.target;
    
    this.setState(prevState => ({
      ...prevState,
      [id]: +value,
    }));
  };
```

Viktor-Kost (changes requested):

Viktor-Kost (changes requested):
.

src/components/Carousel.tsx
```diff
@@ -1,26 +1,71 @@
...
+
+  return (
+    <div className="Carousel">
+      <div className="Carousel-container" style={{ width: carouselWidth }}>
+        <ul
+          className="Carousel__list"
+          style={{
+            transform: `translateX(-${firstIndex * (itemWidth + GAP)}px)`,
+            gap: GAP,
+            transition,
```
SanyaBratashchuk, line 45:
```suggestion
            transition: `transform ${transition}ms ease-in-out`,
```

src/components/Carousel.tsx
```diff
@@ -1,26 +1,71 @@
...
+const Carousel: React.FC<Props> = ({
+  images,
+  frameSize,
+  step,
+  itemWidth,
+  animationDuration,
+}) => {
+  const carouselWidth = frameSize * itemWidth + (frameSize - 1) * GAP;
+  const [firstIndex, setFirstIndex] = useState(0);
+  const [transition, setTransition] = useState('transform 1000ms ease-in-out');
```
SanyaBratashchuk, line 23:
```suggestion
  const [transition, setTransition] = useState(1000);
```

SanyaBratashchuk (changes requested):

maxim2310 (changes requested):

1. firstly would be nice to add some minimum style for this 
2. increasing frame size break carousel
image
3. also would be nice to fix merge conflict 
image

maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

still have this problem 
image

BoGoBooM:
I don't know how I can fix this bug and how can I formulate a request on the platform to get help...
I also saw that other students have this bug

maxim2310 (approved):
So you can skip it but would be nice to fix, you just need to make check if your index + frame size > than quantity of imgs then make index - 1, or you can ask it in fe chat

PR #698, opened 2024-03-21
React_Carousel

Denys209818:
[DEMO LINK](https://denys209818.github.io/react_carousel/)

I have also added an infinite option that can be modifyed in code.

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,26 +1,171 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li>
-        <img src="./img/1.png" alt="1" />
-      </li>
-      <li>
-        <img src="./img/1.png" alt="2" />
-      </li>
-      <li>
-        <img src="./img/1.png" alt="3" />
-      </li>
-      <li>
-        <img src="./img/1.png" alt="4" />
-      </li>
-    </ul>
+interface ImageCollection {
+  images: string[];
+  step: number;
+  frameSize: number;
+  itemWidth: number;
+  animationDuration: number;
+  infinite: boolean;
+}
 
-    <button type="button">Prev</button>
-    <button type="button">Next</button>
-  </div>
-);
+const Carousel: React.FC<ImageCollection> = ({
+  images,
+  step,
+  frameSize,
+  itemWidth,
+  animationDuration,
+  infinite,
+}) => {
+  const [styleList, setStyleList] = useState({
+    transform: 'translateX(0)',
+    transition: `transform ${animationDuration / 1000}s ease-in-out`,
+  });
+
+  const [disabledBtn, setDisabledBtn] = useState({
+    nextBtn: false,
+    prevBtn: true,
+  });
+
+  const returnClickability = () => {
+    if (Object.values(disabledBtn).includes(true)) {
+      setDisabledBtn(() => {
+        return {
+          nextBtn: false,
+          prevBtn: false,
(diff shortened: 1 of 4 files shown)


l4st1m0za (changes requested):
Please provide a demo with all the things asked in the task 
image

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review, this part of the task is incomplete. You need to make a deploy after your changes
image

maxim2310 (changes requested):

1. your frame size is bigger than should
image
2 . if i click back from previous screen, i can't click next, although there are still images there
image
3. item width doesn't change 
4. animation duration same
5. pass all tests before send new review request 
6. also would be nice to center carousel and add border for frame

Denys209818:
I don't know why tests don't pass, but it works as was described

src/components/Carousel.tsx
```diff
@@ -1,26 +1,223 @@
...
+  frameSize,
+  itemWidth,
+  animationDuration,
+  infinite,
+}) => {
+  const [previousWidth, setPreviousWidth] = useState<number>(itemWidth);
+
+  const [styleList, setStyleList] = useState({
+    transform: 'translateX(0)',
+    transition: `transform ${animationDuration / 1000}s ease-in-out`,
```
etojeDenys, line 25: you can just save the numbers here as values of the keys
```suggestion
    transform: 0,
    transition: animationDuration / 1000,
```

src/components/Carousel.tsx
```diff
@@ -1,26 +1,223 @@
...
+    let finalNum = 0;
+
+    if (transformProp.endsWith('px')) {
+      finalNum = parseInt(transformProp.slice(0, -2));
+    } else {
+      finalNum = parseInt(transformProp);
+    }
+
+    return finalNum;
+  };
```
etojeDenys, line 49: after that you can remove this function

src/components/Carousel.tsx
```diff
@@ -1,26 +1,223 @@
...
+  useEffect(() => {
+    const s = getCurrentTransform();
+
+    if (s < 0) {
+      setStyleList({
+        ...styleList,
+        transform: `translateX(${s + itemWidth}px)`,
+      });
+    }
+  }, [frameSize]);
```
etojeDenys, line 171: you can try to combine these useEffects in one

src/components/Carousel.tsx
```diff
@@ -1,26 +1,223 @@
...
+
+  return (
+    <div
+      className="Carousel"
+      style={{
+        maxWidth: `${itemWidth * frameSize}px`,
+      }}
+    >
+      <ul className="Carousel__list" style={styleList}>
```
etojeDenys, line 180: and specify here in style prop the object with our styleList values

src/components/Carousel.tsx
```diff
@@ -1,26 +1,223 @@
...
+
+  const [styleList, setStyleList] = useState({
+    transform: 'translateX(0)',
+    transition: `transform ${animationDuration / 1000}s ease-in-out`,
+  });
+
+  const [disabledBtn, setDisabledBtn] = useState({
+    nextBtn: false,
+    prevBtn: true,
+  });
```
etojeDenys, line 31: looks like it can be the simple var not a state

etojeDenys (changes requested):

src/components/Carousel.tsx
Denys209818: I have tried. It doesn't work

src/components/Carousel.tsx
Denys209818: 3 of 4 useEffects launch changing state and forth of them depends from three useEffects

PR #701, opened 2024-03-24
add solution

IhorRome:
[DEMO LINK](https://IhorRome.github.io/react_carousel/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,24 +3,24 @@
 > [React + Typescript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript)
 
 1. implement a carousel described [here](https://github.com/mate-academy/dom_carousel) as a React component
-    ```jsx harmony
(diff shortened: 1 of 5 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to show the default values to know what and where to enter
image

It will also be better if the clickable area is slightly enlarged
image

l4st1m0za (changes requested):
Please add the inifinite switch to the page
image

IhorRome:
> Please add the inifinite switch to the page image

I can't figure out how to add infinite scroll on "previous"(pointing to the left) button - added only on next button, by adding starting items to the end of array if scrolling over the end

l4st1m0za (approved):
Sorry for confusing you, seems that the task has changed in recent days, and I was not aware for it. You have done good job!

PR #702, opened 2024-03-25
add task solution

ArsenSkandalenko:
[DEMO LINK](https://ArsenSkandalenko.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,6 +4,10 @@ import Carousel from './components/Carousel';
 
 interface State {
   images: string[];
+  step?: number;
+  frameSize?: number;
+  itemWidth?: number;
+  animationDuration?: number;
 }
 
 class App extends React.Component<{}, State> {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -20,17 +24,94 @@ class App extends React.Component<{}, State> {
...
+    const { images, step, frameSize, itemWidth, animationDuration } =
+      this.state;
 
     return (
       <div className="App">
         {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
-
-        <Carousel />
+        <h1 data-cy="title">Carousel with {this.state.images.length} images</h1>
```
l4st1m0za, line 68: You should avoid using 'this.state' inside the render method. Instead, you already destructured the state few lines above so you can use 'images.length' directly.

l4st1m0za (approved):
Good job!

PR #703, opened 2024-03-25
Develop

MariiaSuper:
https://MariiaSuper.github.io/react_carousel/

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,9 @@
-// style go here
+h1 {
+    text-align: center;
+}
+
+.App {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,6 +4,10 @@ import Carousel from './components/Carousel';
 
 interface State {
   images: string[];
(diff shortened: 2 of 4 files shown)


anastasiiavorobiova (changes requested):
Great work! When a user updates the step - the carousel moves unexpectedly. Also, the value here changes when a user clicks on the right or the left arrow

MariiaSuper:
Thanks @anastasiiavorobiova !

What do you mean - if user change step we have to show appropriate image, no? And regards the test `should change scroll text on 1 step`, changing the step should move images into the carousel.
> When a user updates the step - the carousel moves unexpectedly

Looks it the standard behaviour for the input? Is it bad?
> Also, the value here changes when a user clicks on the right or the left arrow

IvanFesenko (changes requested):
- Why does Icon step changes on slide?
- In case slider in ther end add we change Frame side - icons should be aligned to right side
https://github.com/mate-academy/react_carousel/assets/61904995/db2f0f39-5a71-4ffc-9ce0-4dc12a25b3e0

etojeDenys (approved):
good job

PR #707, opened 2024-03-31
Implement carousel

deezhizyu:
[DEMO LINK](https://deezhizyu.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,6 +4,10 @@ import Carousel from './components/Carousel';
 
 interface State {
   images: string[];
+  step: number;
+  frameSize: number;
+  itemWidth: number;
+  animationDuration: number;
 }
 
 class App extends React.Component<{}, State> {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -20,17 +24,83 @@ class App extends React.Component<{}, State> {
...
+        </div>
 
-        <Carousel />
+        <Carousel
+          images={images}
+          step={step}
+          frameSize={frameSize}
+          itemWidth={itemWidth}
+          animationDuration={animationDuration}
+          infinite={true}
```
TarasHoliuk, line 102: it should be `false` by default
also, add a checkbox to have possibility to change it

src/App.tsx
```diff
@@ -20,17 +24,83 @@ class App extends React.Component<{}, State> {
...
+          <label htmlFor="animationDuration">Animation duration</label>
+          <input
+            type="number"
+            id="animationDuration"
+            min="0"
+            max="5000"
+            step="100"
+            value={animationDuration}
+            onChange={event =>
+              this.updateState('animationDuration', +event.target.value)
```
TarasHoliuk, line 91: FYI: you can add `name` attribute to the input same as state key and then get it from the event `event.target.name` (but you will have to use typecasting - to "say" the typescript it's a key of state indeed)

src/components/Carousel.tsx
```diff
@@ -1,26 +1,81 @@
...
+        </ul>
+      </div>
+
+      <button
+        type="button"
+        onClick={() => changeScrollCount(scrollCount - step)}
+      >
+        Prev
+      </button>
+      <button
```
TarasHoliuk, line 70: they should be disabled in case infinity is false and it's end/start of list

TarasHoliuk (changes requested):

TarasHoliuk (approved):
👍

PR #708, opened 2024-04-01
solution

Maksym-Atanasov:
[DEMO LINK](https://Maksym-Atanasov.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -27,10 +27,9 @@ class App extends React.Component<{}, State> {
 
     return (
       <div className="App">
-        {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1 data-cy="title">Carousel with {images.length} images</h1>
 
-        <Carousel />
+        <Carousel images={images} />
       </div>
     );
   }
--- a/src/components/Carousel.scss
+++ b/src/components/Carousel.scss
@@ -1,1 +1,12 @@
-// style go here
+.Carousel {
+
+  &__visible {
+    border: 1px solid red;
+    overflow: hidden;
+  }
+
+  &__list {
+    position: relative;
+    display: flex;
+  }
+}
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,26 +1,79 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li>
-        <img src="./img/1.png" alt="1" />
-      </li>
-      <li>
-        <img src="./img/1.png" alt="2" />
-      </li>
-      <li>
-        <img src="./img/1.png" alt="3" />
-      </li>
-      <li>
-        <img src="./img/1.png" alt="4" />
-      </li>
-    </ul>
+type Props = {
+  images: string[];
+  itemWidth?: number;
+  frameSize?: number;
+  animationDuration?: number;
+  step?: number;
+  infinite?: boolean;
+};
 
-    <button type="button">Prev</button>
-    <button type="button">Next</button>
-  </div>
-);
+const Carousel: React.FC<Props> = ({
+  images,
+  itemWidth = 130,
+  frameSize = 3,
+  step = 3,
+  animationDuration = 1000,
+  infinite = false,
+}) => {
+  const [carouselList, setCarouselList] = useState(0);
+
+  const previous = () =>
+    -carouselList + step < 0
+      ? setCarouselList(carouselList - step)
+      : setCarouselList(0);
+
+  const next = () =>
+    carouselList + step < images.length - step
+      ? setCarouselList(carouselList + step)
+      : setCarouselList(images.length - step);
+
+  if (infinite) {
+    setInterval(() => {
+      if (carouselList + step < images.length - step) {
+        setCarouselList(carouselList + step);
+      } else if (carouselList < images.length - step) {
+        setCarouselList(images.length - step);
+      } else {
+        setCarouselList(0);
+      }
+    }, 3000);
+  }
+
+  return (
+    <div style={{ width: itemWidth * images.length }} className="Carousel">
+      <div
+        style={{ width: itemWidth * frameSize }}
+        className="Carousel__visible"
+      >
+        <ul
+          style={{
+            height: itemWidth,
+            right: carouselList * itemWidth,
+            transition: `all ${animationDuration}ms`,
+          }}
+          className="Carousel__list"
+        >
+          {images.map((image, index) => (
+            <img
+              key={`${index}-${image}`}
+              src={image}
+              alt={String(index)}
+            ></img>
+          ))}
+        </ul>
+      </div>
+
+      <button onClick={previous} data-cy="next" type="button">
+        Prev
+      </button>
+      <button onClick={next} type="button">
+        Next
+      </button>
+    </div>
+  );
+};
 
 export default Carousel;
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,4 @@
-// style go here
+* {
+  margin: 0;
+  padding: 0;
+}


SanyaBratashchuk (changes requested):
image
image

volodymyr-soltys97 (changes requested):
GJ!
To improve:
1. I can't change these fields, so you need to fix it
image

2. Also will be better if you add some styles for these buttons
image

src/components/Carousel.scss
```diff
@@ -1 +1,60 @@
...
+      color: black;
+    }
+  }
+}
+
+    border: 1px solid black;
+    border-radius: 30px;
+    padding: 10px;
+  }
+}
```
DarkMistyRoom: what do these styles apply to?

DarkMistyRoom (changes requested):
- fix `Carousel.scss` code
- there is no possibility of changing Steps
- Frame Size jumps to 50 when I use the arrow up and doesn't change when I use the arrow down
- no need to slide the Frame Size is 10
image

- Infinite mode doesn't work properly 

https://github.com/mate-academy/react_carousel/assets/98764641/c343f0b6-4838-4171-95eb-3b30b2219285

src/components/Carousel.tsx
```diff
@@ -1,26 +1,150 @@
...
+  animationDuration = 1000,
+  infinite = false,
+}) => {
+  const [carouselList, setCarouselList] = useState(0);
+  const [stepState, setStepState] = useState(step);
+  const [frameSizeState, setFrameSizeState] = useState(frameSize);
+  const [itemWidthState, setItemWidthState] = useState(itemWidth);
+  const [animationDurationState, setAnimationDurationState] =
+    useState(animationDuration);
+  const [isInfiniteState, setIsInfiniteState] = useState(infinite);
```
IvanFesenko, line 27: This should be a props, not state

IvanFesenko (changes requested):
- Not fixed
https://github.com/mate-academy/react_carousel/assets/61904995/6f6af998-bc9f-49b3-9e21-2badb046540f
- Infinite - not works
https://github.com/mate-academy/react_carousel/assets/61904995/b7065c56-cbcd-4bda-b9d2-9214b66962cc
- Check this case
https://github.com/mate-academy/react_carousel/assets/61904995/6c86313b-c96f-4bc4-ab75-c0ff2278f649

src/components/Carousel.tsx
Maksym-Atanasov: it will not work because its rendering on page

sTorba24 (changes requested):
Pay attention to fix ALL comments before request review
If you have any issues with it - feel free to ask in fe_chat

it scrolls only 2 elements with such settings, but should be 8 or how much is possible
image

inifinite mode still doesn't work

src/components/Carousel.tsx
TarasHoliuk: You should have these values in the state in App.tsx, and pass them here as props
Props change also cause re-render

In Carousel.tsx should be only a carousel with buttons. Inputs and their values shouldn't be here

src/components/Carousel.tsx
```diff
@@ -1,26 +1,168 @@
...
-      <li>
-        <img src="./img/1.png" alt="4" />
-      </li>
-    </ul>
+type Props = {
+  images: string[];
+  itemWidth?: number;
+  frameSize?: number;
+  step?: number;
+  animationDuration?: number;
```
TarasHoliuk, line 9:
```suggestion
  itemWidth: number;
  frameSize: number;
  step: number;
  animationDuration: number;
```

src/components/Carousel.tsx
```diff
@@ -1,26 +1,168 @@
...
+  const [stepState, setStepState] = useState(step);
+  const [frameSizeState, setFrameSizeState] = useState(frameSize);
+  const [itemWidthState, setItemWidthState] = useState(itemWidth);
+  const [animationDurationState, setAnimationDurationState] =
+    useState(animationDuration);
+  const [isInfiniteState, setIsInfiniteState] = useState(infinite);
+
+  const previous = () => {
+    if (-carouselList + stepState < 0) {
+      setCarouselList(carouselList - stepState);
```
TarasHoliuk, line 31:
```suggestion
    if (carouselList - stepState < 0) {
      setCarouselList(0);
```

If the expected value is out of range just set 0 in `previous`. in `next` - the maximum value

src/components/Carousel.tsx
```diff
@@ -1,26 +1,168 @@
...
+    } else {
+      if (carouselList === 0 && infinite) {
+        setCarouselList(images.length - frameSizeState);
```
TarasHoliuk, line 34: this logic doesn't work as expected

Let's define how should it work. In case the expected position (`carouselList`) is out of range _always_ (no matter if `infinite` is true or false) scroll to the start/end (depends on the button). But if it's true and you are already on the start/end - scroll images.

Fro example if you on the start, step = 3, frameSize = 3, infinite = true and you press `prev` button scroll to the last 3 images

src/components/Carousel.tsx
```diff
@@ -1,26 +1,168 @@
...
+          disabled={isInfiniteState === false && carouselList === 0}
+          type="button"
+        >
+          Prev
+        </button>
+        <button
+          className="Carousel__button"
+          onClick={() => next()}
+          disabled={
+            isInfiniteState === false && carouselList === images.length - 1
```
TarasHoliuk, line 90: would be nice to add some additional style when button disabled to show it

TarasHoliuk (changes requested):
Hi!
Some logic still doesn't work properly
I've added comments to describe how should it work. Hope it will help. But in cease you have more questions - ask them in the chat or join to the Q&A session

src/components/Carousel.tsx
Maksym-Atanasov: if I on last smile and when I press prev, it returns to the first, I think it didn't need to work so

src/components/Carousel.tsx
SanyaBratashchuk: not fixed. Would be better to move these values in state in App.tsx. If you have any questions feel free to ask in fe_chat.

src/components/Carousel.tsx
```diff
@@ -1,26 +1,169 @@
...
+          type="number"
+          min={0}
+          step={100}
+          onChange={e => setAnimationDurationState(Number(e.target.value))}
+          value={animationDurationState}
+        />
+        <p>Infinite</p>
+        <input
+          className="Carousel__input"
+          checked={isInfiniteState ? true : false}
```
SanyaBratashchuk, line 160:
```suggestion
          checked={isInfiniteState}
```

SanyaBratashchuk (changes requested):

vadiimvooo (changes requested):
Good job, but there is one problem exists:
When you selecting 5 or more elements and press next button is scrolls untill the end but shouldn't do it. I've attached the video of bug. Please feel free to ask for help in fe_chat or visit the QnA session.
https://www.loom.com/share/45baa4bb52c54290b9944e646a69c1a3

src/App.tsx
```diff
@@ -20,17 +32,64 @@ class App extends React.Component<{}, State> {
...
+          stepState={stepState}
+          animationDurationState={animationDurationState}
+          isInfiniteState={isInfiniteState}
+          images={images}
+          setAnimationDurationState={this.setAnimationDurationState}
+          setFrameSizeState={this.setFrameSizeState}
+          setItemWidthState={this.setItemWidthState}
+          setStepState={this.setStepState}
+          setIsInfiniteState={this.setIsInfiniteState}
+        />
```
maxim2310, line 92: carousel must accept only these props, because carousel must be dumm component 
<Carousel
  images={['url1', 'url2']}
  step={3}
  frameSize={3}
  itemWidth={130}
  animationDuration={1000}
  infinite={false}
/>

all inputs need to placed in app component and add id attribute for them due the test

maxim2310 (changes requested):
Try to refactor a bit and pass all tests

src/App.tsx
```diff
@@ -24,13 +31,27 @@ class App extends React.Component<{}, State> {
...
-        <Carousel />
+        <Carousel
+          images={images}
+          step={step}
+          frameSize={frameSize}
+          itemWidth={itemWidth}
+          animationDuration={animationDuration}
+          infinite={false}
+        />
       </div>
```
etojeDenys: image

you should move all inputs as the mentor said and logic with them to the App.tsx and just pass their values to the Carousel

```
<div className="App">
 ...your inputs
 <Carousel  ... />
</div>
```

src/components/Carousel.tsx
```diff
@@ -1,26 +1,152 @@
...
+        if (
+          stepState > frameSizeState &&
+          carouselList === images.length - stepState
+        ) {
+          setCarouselList(images.length - frameSizeState);
+        } else {
+          if (
+            stepState > images.length / 2 &&
+            carouselList > images.length / 2
+          ) {
```
etojeDenys, line 63: it looks too complicated, please, feel free to ask for some help in the chat with that to simplify it

src/components/Carousel.tsx
```diff
@@ -1,26 +1,152 @@
...
+            transition: `all ${animationDurationState}ms`,
+          }}
+          className="Carousel__list"
+        >
+          {images.map((image, index) => (
+            <img
+              key={`${index}-${image}`}
+              src={image}
+              alt={String(index)}
+            ></img>
```
etojeDenys, line 121:
```suggestion
            <img
              key={`${index}-${image}`}
              src={image}
              alt={String(index)}
            />
```

src/components/Carousel.tsx
```diff
@@ -1,26 +1,152 @@
...
+          ))}
+        </ul>
+      </div>
+
+      <div className="Carousel__buttons">
+        <button
+          className="Carousel__button"
+          onClick={() => previous()}
```
etojeDenys, line 129:
```suggestion
          onClick={previous}
```
we can simplify it

etojeDenys (changes requested):

DarkMistyRoom (changes requested):
Infinite mode works only for the left direction

lerastarynets (approved):
You've got this, keep pushing forward!

PR #710, opened 2024-04-02
Solution

VictoriaPukalo:
[DEMO LINK](https://VictoriaPukalo.github.io/react_carousel/)


SanyaBratashchuk (changes requested):
It doesn't work as expected. 
I can't see any pictures except the first three after scrolling. 
image

vadiimvooo (changes requested):
Good job. There is also one bug with scrolling the images in infinite mod (the next button is not working for some reason)
https://www.loom.com/share/3e96ad7f730b464d80fda3ce675e2fcc

PR #711, opened 2024-04-03
solution

yuliia-nudyk:
DemoLink - https://yuliia-nudyk.github.io/react_carousel/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The transition doesn't work
image

Also you need to show default values in inputs

etojeDenys (approved):
lgtm

PR #719, opened 2024-04-10
have problems :(

KatsuboDmytro:
https://KatsuboDmytro.github.io/react_carousel/

1. how to make infinite array?
2. how to fix this problems?
    1) should display 3 images by default
    2) should scroll 3 images by default
    3) should change scroll text on 1 step
    4) should change width of image
    5) should change frame size (I have itemWidth={imageWidth} yet. frameSize needn't for this situstion


volodymyr-soltys97 (changes requested):
Hi, you need to add inputs on the page, which will have such default values and the user must be able to change this data in the input
image

l4st1m0za (approved):
Good job!

PR #728, opened 2024-04-24
task done

Reaffith:
[DEMO LINK](https://Reaffith.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -70,6 +70,7 @@ const Carousel: React.FC<{
       </div>
 
       <button
+        data-cy="next"
         type="button"
         onClick={() => showNext()}
         disabled={currentIndex + frameSize === images.length}
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,12 @@
-// style go here
+form{
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  gap: 20px;
+}
+
+label {
+  font-family: sans-serif;
+  font-weight: 600;
+}


src/App.tsx
```diff
@@ -30,7 +30,13 @@ class App extends React.Component<{}, State> {
         {/* eslint-disable-next-line */}
         <h1>Carousel with {images.length} images</h1>
 
-        <Carousel />
+        <Carousel
+          images={images}
+          animationDuration={1000}
+          step={3}
+          itemWidth={130}
+          frameSize={3}
```
TarasHoliuk, line 38: it all should be dynamic state values, not hardcoded 1000 or 3, etc.

Below Carousel, you should render inputs, where you change these values

src/components/Carousel.tsx
```diff
@@ -1,26 +1,89 @@
...
+  itemWidth: number;
+}> = ({
+  images,
+  animationDuration = 1000,
+  itemWidth = 130,
+  step = 3,
+  frameSize = 3,
+}) => {
+  const [translate, setTranslate] = useState(0);
+  const [currentStep, setCurrentStep] = useState(1);
```
TarasHoliuk, line 18: it's not step
currentImageIndex or offset, etc.

src/components/Carousel.tsx
```diff
@@ -1,26 +1,89 @@
...
+  frameSize: number;
+  itemWidth: number;
+}> = ({
+  images,
+  animationDuration = 1000,
+  itemWidth = 130,
+  step = 3,
+  frameSize = 3,
+}) => {
+  const [translate, setTranslate] = useState(0);
```
TarasHoliuk, line 17: you don't need both - `translate` and `currentStep `. You need one state to store the index of first image you display

TarasHoliuk: Then compute translate and save it just as const, _example_:
`const translate = currentState * itemWidth`

src/components/Carousel.tsx
```diff
@@ -1,26 +1,89 @@
...
+  const [currentStep, setCurrentStep] = useState(1);
+  const lengthOfList = images.length * itemWidth;
+  const lengthOfWrap = frameSize * itemWidth;
+  const stepLength = step * itemWidth;
+  const maxFullSteps = Math.floor(lengthOfList / stepLength);
```
TarasHoliuk, line 22: why such formula? looks wrong - check it after adding inputs (possibility to change step, item width, etc)

src/components/Carousel.tsx
```diff
@@ -1,26 +1,89 @@
...
+              }}
+            >
+              <img src={item} alt={item} />
+            </li>
+          ))}
+        </ul>
+      </div>
+
+      <div className="cover">
+        <button type="button" onClick={() => showPrevious()}>
```
TarasHoliuk, line 78: should be disabled if it's at the start of the images list. Similar with next button

TarasHoliuk (changes requested):
Carefully read the task description. Ask in the chat if you need help

l4st1m0za (changes requested):
1) When I change the width, your carousel gets shifted
image

if i will set frame size as 5 in this case, i won't be able to to press **Next** button
image

2) At this point, I can't scroll further, but the next button, still has styles as if i can press it
image

3) If I will set step as 5, in frame size 3, then, my elements won't be shown at the last step
image

PR #729, opened 2024-04-25
adding task solution

eduvinagre:
[DEMO LINK](https://eduvinagre.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,119 @@
-import React from 'react';
-import './App.scss';
+import React, { useState } from 'react';
 import Carousel from './components/Carousel';
+import './App.scss';
 
-interface State {
-  images: string[];
-}
+const images = [
+  './img/1.png',
+  './img/2.png',
+  './img/3.png',
+  './img/4.png',
+  './img/5.png',
+  './img/6.png',
+  './img/7.png',
+  './img/8.png',
+  './img/9.png',
+  './img/10.png',
+];
 
-class App extends React.Component<{}, State> {
-  state = {
-    images: [
-      './img/1.png',
-      './img/2.png',
-      './img/3.png',
-      './img/4.png',
-      './img/5.png',
-      './img/6.png',
-      './img/7.png',
-      './img/8.png',
-      './img/9.png',
-      './img/10.png',
-    ],
+const initialOptions = {
+  itemWidth: 130,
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,119 @@
...
-        {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <label className="App__label" htmlFor="stepId">
+          Step:
+          <input
+            type="number"
+            className="App__input"
+            id="stepId"
+            name="step"
+            min={initialOptions.step}
```
pedro-ruas, line 82: Why are you setting a min value based on step? What if the user wants to move 1 image at a time...
I don't think there is a need for this min here, or min should be 1 if needed

src/App.tsx
```diff
@@ -1,39 +1,119 @@
...
-  }
-}
+        <label className="App__label" htmlFor="animationDurationId">
+          Animation Duration:
+          <input
+            type="number"
+            className="App__input"
+            id="animationDurationId"
+            name="animationDuration"
+            min={initialOptions.animationDuration}
```
pedro-ruas, line 96: Same here, I don't think a min value based on the initial one is needed, it just has to be above 0

src/App.tsx
```diff
@@ -1,39 +1,119 @@
...
+        </label>
+
+        <label className="App__label" htmlFor="frameId">
+          Frame Size:
+          <input
+            type="number"
+            className="App__input"
+            id="frameId"
+            name="frameSize"
+            min={initialOptions.frameSize}
```
pedro-ruas, line 68: Same here, why the min value?

src/App.tsx
```diff
@@ -1,39 +1,119 @@
...
+
+      <form className="App__form" action="#" method="post">
+        <label className="App__label" htmlFor="itemId">
+          Item Width:
+          <input
+            type="number"
+            className="App__input"
+            id="itemId"
+            name="itemWidth"
+            min={initialOptions.itemWidth}
```
pedro-ruas, line 54: And here

pedro-ruas (changes requested):
Minimal UX improvements

EdPirro (approved):
Great job, @eduvinagre! Keep it up! 🚀

PR #732, opened 2024-04-30
add solution

reznik-denis:
[DEMO LINK](https://reznik-denis.github.io/react_carousel/)


etojeDenys (changes requested):
let's fix this empty space
image

reznik-denis:
I fix this problem when i chenge value to defaultValue, and I fixed problem with test, but I had to relocate button block under carousel. When I fix it on same level cupress think it another image and tests unchecked.

etojeDenys (changes requested):
maybe you forgot to redeploy? cause now nothing has changed, I still can reproduce the bug

reznik-denis:
at start i don't understand where is problem in app, but now i understend. So, I can't chenge form in Carousel component, so I will have to finde another desigion, how control the frameSize inside Carousel. Better that i can find it's use hook useState whit useEffect and change transform every time when we change frameSize

etojeDenys (approved):
good job

PR #736, opened 2024-05-04
Solution

SlavaRevit:
- [DEMO LINK](https://slavarevit.github.io/react_carousel/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,9 @@
-// style go here
+.App {
+  display: flex;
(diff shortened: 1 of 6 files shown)


lerastarynets (changes requested):
These fields should be filled by default also pls write INIFINITE near the checkbox so that it's purpose is clear for user

Screenshot at May 06 10-26-07

l4st1m0za (approved):
Good job!

PR #738, opened 2024-05-08
add task solution

OleksandraKoshyk:
[DEMO LINK](https://OleksandraKoshyk.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,26 +1,85 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { Props } from '../types/Props';
 import './Carousel.scss';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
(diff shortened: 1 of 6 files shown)


src/components/Carousel.tsx
```diff
@@ -1,26 +1,85 @@
...
+      <div
+        className="Carousel__container"
+        style={{
+          width: `${itemWidth * frameSize}px`,
+        }}
+      >
+        <ul
+          className={`Carousel__list ${infinite ? 'Carousel__list--animated' : ''}`}
+          style={{
+            transition: `all 100ms ease`,
```
TarasHoliuk, line 47: The idea was to use the transition `input` value here.

Infinity `true` should have allowed to click `next` button when the last images are shown. In this case, the next button should scroll images to start of the list.

But you did similar things, and it looks good 👍 All tests passed - approved ✔️

TarasHoliuk (approved):

PR #747, opened 2024-06-04
Develop

aMtiIV:
[DEMO LINK](https://aMtiIV.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -45,10 +45,15 @@ class App extends React.Component<{}, State> {
   };
 
   handleFrameSizeChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    const newFrameSize = parseInt(event.target.value);
-    const fullSize = this.state.images.length;
+    let newFrameSize = parseInt(event.target.value);
+    const { images, infinite } = this.state;
+    const listSize = images.length;
 
-    if (this.isNumberPositive(newFrameSize) && newFrameSize <= fullSize) {
+    if (this.isNumberPositive(newFrameSize)) {
+      if (!infinite && newFrameSize > listSize) {
+        newFrameSize = listSize;
+      }
+
       this.setState({ frameSize: newFrameSize });
     }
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -20,17 +25,150 @@ class App extends React.Component<{}, State> {
...
+  ) => {
+    const newAnimationDuration = parseInt(event.target.value);
+
+    if (this.isNumberPositive(newAnimationDuration)) {
+      this.setState({ animationDuration: newAnimationDuration });
+    }
+  };
+
+  handleInfiniteChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { images, infinite } = this.state;
```
BogdanMaliuta, line 80: you always doing this in your functions, mb better option to destruct its you cimponent?

src/components/Carousel.tsx
```diff
@@ -1,26 +1,237 @@
...
+  render(): React.ReactNode {
+    const { position, duringAnimation } = this.state;
+
+    return (
+      <div className="Carousel">
+        <button
+          className={`Carousel__button Carousel__button--prev${
+            this.isPositionBeforeRange(position - 1)
+              ? ' Carousel__button--disabled'
+              : ''
```
BogdanMaliuta, line 210: don`t need to use here empty option with ternary operator, you can do this like this: classname: condition

BogdanMaliuta (changes requested):
Nice work!
Some fixes up to you, also you can change if you want your styles, make your inputs in column, add margins for better look

src/App.tsx
aMtiIV: @BogdanMaliuta Could you explain more?

BogdanMaliuta: look at lines 49,80,92 its code duplication.

BogdanMaliuta (approved):
Nice job, but remove code duplications

PR #751, opened 2024-06-13
add carousel

ly-manka:
[DEMO LINK](https://ly-manka.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,14 @@
 import React from 'react';
 import './App.scss';
-import Carousel from './components/Carousel';
+import { Carousel } from './components/Carousel';
 
 interface State {
   images: string[];
+  step: number;
+  frameSize: number;
+  itemWidth: number;
+  animationDuration: number;
+  infinite: boolean;
 }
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -20,17 +25,117 @@ class App extends React.Component<{}, State> {
...
+
+        <div className="App__fields">
+          <div className="wrapper">
+            <label htmlFor="step">Step</label>
+            <input
+              id="step"
+              className="wrapper__field"
+              type="number"
+              min={1}
+              max={10}
```
DarkMistyRoom, line 74: would be better to move all default numbers to constants

DarkMistyRoom (changes requested):
Good job! But the infinite mode is not about autoscrolling. It should make it possible to scroll to the first item when we click `right` after reaching the end (analogical for the left side)

volodymyr-soltys97 (changes requested):
GJ!
To improve
I can't change item width for less than now, also fix it in the animation duration 
image
image

VitaliyBondarenko1982 (approved):
Looks good

PR #754, opened 2024-06-21
add task solution

kravets111:
[DEMO LINK](https://kravets111.github.io/react_carousel/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to disable next button when infinite mode is turn off 
image

Fix the container width 
image

vadiimvooo (approved):
Well done

PR #759, opened 2024-06-29
solution

getchards:
[DEMO LINK](https://getchards.github.io/react_carousel/)


src/components/Carousel.tsx
```diff
@@ -1,26 +1,101 @@
...
+  return (
+    <div className="Carousel">
+      <div
+        className="Carousel__container"
+        style={{
+          width: `${itemWidth * frameSize}px`,
+        }}
+      >
+        <ul
+          className={`Carousel__list ${infinite ? 'Carousel__list--animated' : ''}`}
```
SanyaBratashchuk, line 61: it is better to use libraries like `classnames` or `clsx` for conditional classnames

SanyaBratashchuk (approved):

PR #762, opened 2024-07-04
Develop

Mariana-VV:
[DEMO LINK](https://Mariana-VV.github.io/react_carousel/)


anastasiiavorobiova (changes requested):
Great start! Currently, the carousel doesn't work if a user changes settings. Also, images disappear after the next/prev button click.

src/components/Carousel.tsx
```diff
@@ -1,18 +1,91 @@
...
+        <button
+          type="button"
+          onClick={() => {
+            if (position >= 0 || position < maxPosition) {
+              setIndex(prevTranslate => prevTranslate
+                - data.step * data.itemWidth - 5 * data.step);
+              setPositon(prevPosition => prevPosition
+                - data.step * data.frameSize);
+            }
+          }}
```
VitaliyBondarenko1982, line 62: it is bad practice to write functions in jsx.
make handler in component and pass reference here/
same for next button.

src/components/CarouselForm/CarouselForm.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+type Props = {
+  onSubmit: React.Dispatch<React.SetStateAction<Data>>;
+};
+
+export const CarouselForm: React.FC<Props> = ({ onSubmit }) => {
+  const [data, setData] = useState<Data>({
+    frameSize: 3,
+    itemWidth: 130,
+    step: 3,
```
VitaliyBondarenko1982, line 12: also should be transition value

VitaliyBondarenko1982 (changes requested):
Good job.
Needs some improvements.
- not fixed prev comment. in some cases images dissapears
carousel
- some unexpected view which not which not matches values. always should be visible full image.
Screenshot 2024-07-06 at 08 41 26
- add labels for inputs.

volodymyr-soltys97 (changes requested):
Hi, I can't change transition time. Need to fix it
image

It will be better if these buttons are in the center
image

volodymyr-soltys97 (approved):
Great work 🔥

PR #769, opened 2024-07-29
Carousel

AvramenkoMarina:
[DEMO LINK](https://AvramenkoMarina.github.io/react_carousel/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,39 @@
-// style go here
+iframe {
+  display: none;
+}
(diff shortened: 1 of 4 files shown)


etojeDenys (approved):
let's fix this case

https://github.com/user-attachments/assets/422df18d-ad94-4573-8177-9b47a2ce1450

etojeDenys (changes requested):
better to fix it, thx

PR #772, opened 2024-08-12
Solution

Teollan:
[DEMO LINK](https://Teollan.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,6 +4,11 @@ import Carousel from './components/Carousel';
 
 interface State {
   images: string[];
+  itemWidth: number;
+  frameSize: number;
+  step: number;
+  animationDuration: number;
+  isInfinite: boolean;
 }
 
 class App extends React.Component<{}, State> {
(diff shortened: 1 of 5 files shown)


AndriiZakharenko (approved):
good job!

src/App.tsx
```diff
@@ -20,17 +25,87 @@ class App extends React.Component<{}, State> {
...
-        {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1 data-cy="title">Carousel with {this.state.images.length} images</h1>
+
+        <>
+          <label htmlFor="itemId">Item width: </label>
+          <input
+            onChange={event =>
+              this.setState({ itemWidth: +event.target.value })
+            }
```
BudnikOleksii, line 46: Create handlers instead of passing anonim functions to `onChange`

src/App.tsx
```diff
@@ -20,17 +25,87 @@ class App extends React.Component<{}, State> {
...
+            type="number"
+            name="itemId"
+            id="itemId"
+            defaultValue={130}
```
BudnikOleksii, line 50: Why do you use `defaultValue`? The value should depend on the state, and these values are magic values, create a constant

src/components/Carousel.tsx
```diff
@@ -1,26 +1,142 @@
...
+  animationDuration?: number;
+  infinite?: boolean;
+};
+
+const Carousel: React.FC<Props> = ({
+  images,
+  step = 3,
+  frameSize = 3,
+  itemWidth = 130,
+  animationDuration = 1000,
```
BudnikOleksii, line 18: That's what I meant, you use these value in 3 places, create a constant so it will be the same in any case

src/components/Carousel.tsx
```diff
@@ -1,26 +1,142 @@
...
+  infinite = false,
+}) => {
+  const [index, setIndex] = useState(0);
```
BudnikOleksii, line 21: What index is it?

src/components/Carousel.tsx
```diff
@@ -1,26 +1,142 @@
...
+  const [isPrevPossible, setIsPrevPossible] = useState(infinite || index > 0);
+  const [isNextPossible, setIsNextPossible] = useState(
+    infinite || index < images.length - frameSize,
+  );
```
BudnikOleksii, line 25: These values are dynamic, don't store them in the state, just create variables and do calculations

src/components/Carousel.tsx
```diff
@@ -1,26 +1,142 @@
...
+
+  const containerStyle = { width: `${containerWidth}px` };
+  const listStyle = {
+    left: `-${offset}px`,
+    transitionDuration: `${animationDuration}ms`,
+  };
+
+  const imageHiddenStyle = {
+    visibility: 'hidden',
+    transitionDelay: `${animationDuration}ms`, // creates an illusion of smooth motion without angering the cypress gods
```
BudnikOleksii, line 38: Remove all comments

src/components/Carousel.tsx
```diff
@@ -1,26 +1,142 @@
...
+        return;
+      }
+
+      setIsPrevPossible(newIndex > 0);
+      setIsNextPossible(newIndex < images.length - frameSize);
+    },
+    [frameSize, images.length, infinite],
+  );
+
+  const goToNextIndex = useCallback(() => {
```
BudnikOleksii, line 53: It doesn't make sense to wrap these functions with `useCallback` as you don't pass them to other components

src/components/Carousel.tsx
```diff
@@ -1,26 +1,142 @@
...
+    setIndex(oldIndex => {
+      let newIndex = oldIndex + step;
+
+      if (newIndex >= images.length - (frameSize - step)) {
+        newIndex = 0;
+      } else if (newIndex > images.length - frameSize) {
+        newIndex = images.length - frameSize;
+      }
+
+      updateMovement(newIndex);
```
BudnikOleksii, line 63: I don't get what the point of these functions, but at least in the `setState` calls it's not a good idea to call other side functions

BudnikOleksii (changes requested):
Good work 👍 
Left some comments, check them. Also, it would be great to add some validations, as now user can set any values and they can be invalid
image

src/components/Carousel.tsx
Teollan: Yep, you are right. I don't know what i was thinking when I wrote this 😅

src/components/Carousel.tsx
Teollan: The functions are supposed to calculate a correct index in some edge cases, like when `step !== frameSize` or when number of pictures isn't perfectly divisible by `frameSize`. And they work both for normal and infinite `Carousel`.

TLDR; it's a mess, but it's functional)

BudnikOleksii (approved):
👍

PR #773, opened 2024-08-12
add task solution

AndriiZakharenko:
[DEMO LINK](https://AndriiZakharenko.github.io/react_carousel/)

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,4 +1,5 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
+import classNames from 'classnames';
 import { Props } from '../types/Props';
 import './Carousel.scss';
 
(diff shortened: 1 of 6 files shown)


src/components/Carousel.tsx
```diff
@@ -1,26 +1,85 @@
...
+    setCurrentShift(maxShift);
+  }
+
+  const fullShift = `-${itemWidth * maxShift}px`;
+
+  document.documentElement.style.setProperty('--shift', fullShift);
+  document.documentElement.style.setProperty(
+    '--animation-duration',
+    `${animationDuration}ms`,
+  );
```
BudnikOleksii, line 26: It's a bad idea to set styles/css-variables directly like this
If you still need it at least do it in `useEffect`

src/components/Carousel.tsx
```diff
@@ -1,26 +1,85 @@
...
+      >
+        <ul
+          className={`Carousel__list ${infinite ? 'Carousel__list--animated' : ''}`}
+          style={{
+            transition: `all 100ms ease`,
+            transform: `translateX(-${currentShift * itemWidth}px)`,
+          }}
+        >
+          {images.map((image, index) => (
+            <li className="Carousel__item" key={index}>
```
BudnikOleksii, line 52: Don't use index as a key

src/components/Carousel.tsx
```diff
@@ -1,26 +1,85 @@
...
+  return (
+    <div className="Carousel">
+      <div
+        className="Carousel__container"
+        style={{
+          width: `${itemWidth * frameSize}px`,
+        }}
+      >
+        <ul
+          className={`Carousel__list ${infinite ? 'Carousel__list--animated' : ''}`}
```
BudnikOleksii, line 45: Use classnames library to set classes conditionally

src/components/Carousel.tsx
```diff
@@ -1,26 +1,85 @@
...
+          style={{
+            transition: `all 100ms ease`,
```
BudnikOleksii, line 47: Specify to which properties should transition apply

src/components/Carousel.tsx
```diff
@@ -1,26 +1,85 @@
...
+          ))}
+        </ul>
+      </div>
+
+      <div className="Carousel__button-group">
+        <button
+          type="button"
+          className="Carousel__button"
+          onClick={showPrev}
+          disabled={infinite}
```
BudnikOleksii, line 69: It should be disabled if it's not infinite and when it's the first slide so you can't move to the previous one

src/components/Carousel.tsx
```diff
@@ -1,26 +1,85 @@
...
+        >
+          Prev
+        </button>
+        <button
+          type="button"
+          className="Carousel__button"
+          onClick={showNext}
+          disabled={infinite}
```
BudnikOleksii, line 77: The same but in case of the last items.
Also, infinite behavior isn't correct. Infinite means that you can scroll the carousel infinitely, when it's the last slide at the next step you show from the beginning and so on.
Simple example how the infinite should work: https://codesandbox.io/p/sandbox/swiper-infinite-loop-bcp5b

BudnikOleksii (changes requested):
Good work 👍 
Left some suggestions, also it would be great if you add some validations so a user can't do something like this easily
image

mrkovaliv (approved):

PR #774, opened 2024-08-15
add App Carousel

VvynnykV:
[DEMO LINK](https://VvynnykV.github.io/react_carousel/)


src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,82 @@
...
+  };
+
+  return (
+    <div
+      className="Carousel"
+      style={{ width: `${itemWidth * frameSize - 1}px` }}
+    >
+      <ul className="Carousel__list" style={styleList}>
+        {images.map((image, i) => (
+          <li key={i}>
```
anastasiia-tilikina, line 54: don't use index as key

anastasiia-tilikina (changes requested):
GJ

we need to make some fixes
pls check my comment and watch [this](https://www.loom.com/share/0d17b320f0d44de098f8a558e590c97b?sid=31abb18c-10f1-4b17-b760-968f5c6550c6)

src/components/Carousel/Carousel.tsx
VvynnykV: I dont have such issues with Frame size that in your video-message, also I have fixed key

volodymyr-soltys97 (approved):
Great work! 🔥

PR #775, opened 2024-08-19
Develop

KrisPetrienko:
[DEMO LINK](https://KrisPetrienko.github.io/react_carousel/)


lerastarynets (changes requested):
You should have 10 smiles instead of 6 in your carousel, also infinite mode and changing step size doesn't seem to be working now, pls fix

PR #776, opened 2024-08-26
Solution

vladkorobka:
demo: https://vladkorobka.github.io/react_carousel/

### DIFF
--- a/src/components/Carousel.tsx
+++ b/src/components/Carousel.tsx
@@ -1,26 +1,103 @@
 import React from 'react';
 import './Carousel.scss';
+import classNames from 'classnames';
 
-const Carousel: React.FC = () => (
-  <div className="Carousel">
-    <ul className="Carousel__list">
-      <li>
-        <img src="./img/1.png" alt="1" />
-      </li>
-      <li>
-        <img src="./img/1.png" alt="2" />
-      </li>
-      <li>
-        <img src="./img/1.png" alt="3" />
-      </li>
-      <li>
-        <img src="./img/1.png" alt="4" />
-      </li>
-    </ul>
+type Props = {
+  images: string[];
+  step: number;
+  frameSize: number;
+  itemWidth: number;
+  animationDuration: number;
+  infinite?: boolean;
+};
 
-    <button type="button">Prev</button>
-    <button type="button">Next</button>
-  </div>
-);
+const Carousel: React.FC<Props> = ({
+  images,
+  step,
+  frameSize,
+  itemWidth,
+  animationDuration,
+  infinite,
+}) => {
+  const [slide, setSlide] = React.useState(0);
+
+  const totalItems = images.length;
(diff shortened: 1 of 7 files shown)


volodymyr-soltys97 (changes requested):
Hi, you need to fix all failed tests, if you need help feel free in the fe_chat 
image

src/components/Carousel.tsx
```diff
@@ -1,26 +1,103 @@
...
+          ))}
+        </ul>
+      </div>
+
+      <div className="Carousel__navigate">
+        <button
+          className={classNames([
+            'Carousel__button',
+            !previousSlide && !infinite ? 'Carousel__button--disable' : '',
+          ])}
```
TarasHoliuk, line 78:
```suggestion
          className={classNames('Carousel__button', {
            'Carousel__button--disable': !previousSlide && !infinite,
          })}
```
correct classNames usage in this case

src/components/Carousel.tsx
```diff
@@ -1,26 +1,103 @@
...
+  const handleNextClick = () => {
+    if (nextSlide) {
+      setSlide(currentSlide => (currentSlide + 1) % (slideItems + 1));
+    }
+  };
+
+  const handlePrevClick = () => {
+    if (previousSlide) {
+      setSlide(currentSlide =>
+        currentSlide === 0 ? slideItems : currentSlide - 1,
```
TarasHoliuk, line 56:
```suggestion
        currentSlide === 0 ? slideItems : currentSlide - step,
```
logic is not clear to me: you should set correct slide taking into account `step` when click and do not use it when calculate translateX

Now, when I select step = 5, scroll to the end and then click `prev` button it scrolls 2 items instead of 5

vladkorobka: you're right, thanks

vladkorobka:
image

src/components/Carousel.tsx
```diff
@@ -1,26 +1,101 @@
...
+  const containerStyle = React.useMemo(
+    () => ({
+      width: `${itemWidth * frameSize}px`,
+    }),
+    [itemWidth, frameSize],
+  );
+
+  const listStyle = React.useMemo(() => {
+    const maxTranslateX = (totalItems - frameSize) * itemWidth;
+    const translateX = Math.min(slide * itemWidth * step, maxTranslateX);
```
TarasHoliuk, line 39:
```suggestion
    const translateX = Math.min(slide * itemWidth, maxTranslateX);
```
Now it scrolls even when I just change step and do not touch the scroll buttons (if the middle of the images list is displayed at the moment - i.e. slide !== 0).

Also, it scrolls too much, at least on Prev. button click

TarasHoliuk (changes requested):
Try different variants - change input values, click next and prev buttons with different input values, etc. Make sure logic works as expected)

anastasiiavorobiova (approved):
Great work!

PR #780, opened 2024-09-12
add solution

Valentyn-Radobenko:
[DEMO LINK](https://valentyn-radobenko.github.io/react_carousel/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,91 @@
-// style go here
+@import './reset';
+
+.App {
+  box-sizing: border-box;
+  height: 100vh;
+  padding: 20px;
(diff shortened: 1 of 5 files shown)


Valentyn-Radobenko:
Tests can't be passed because it said that the page should have picture size 130 px by default and the page should change width of image. But the picture size is clearly 130px and the page can change width of image.
image

volodymyr-soltys97 (changes requested):
Hi, try to run this command `CYPRESS_VERIFY_TIMEOUT=1000000 npm i` for fix failed tests, if you need help feel free in the fe_chat

volodymyr-soltys97 (approved):
Great! 🔥

PR #781, opened 2024-09-17
Develop

przwojwwp:
[DEMO LINK](https://przwojwwp.github.io/react_carousel/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,7 +21,7 @@ export const App: React.FC = () => {
   const [itemWidth, setItemWidth] = useState(130);
   const [frameSize, setFrameSize] = useState(3);
   const [animationDuration, setAnimationDuration] = useState(1000);
-  const [infinite, setInfinite] = useState(false);
+  const [infinite, setInfinite] = useState(true);
 
   const handleSetStep = (event: React.ChangeEvent<HTMLInputElement>) => {
     setStep(Math.round(Number(event.target.value)));
@@ -74,7 +74,11 @@ export const App: React.FC = () => {
         onChange={handleAnimationDuration}
       />
 
-      <button type="button" onClick={handleToggleInfinite}>
+      <button
+        className="toggleButton"
+        type="button"
+        onClick={handleToggleInfinite}
+      >
         Toogle Infinite
       </button>
       {/* Przekazanie images jako props do Carousel */}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,6 +21,7 @@ export const App: React.FC = () => {
   const [itemWidth, setItemWidth] = useState(130);
   const [frameSize, setFrameSize] = useState(3);
   const [animationDuration, setAnimationDuration] = useState(1000);
+  const [infinite, setInfinite] = useState(false);
 
   const handleSetStep = (event: React.ChangeEvent<HTMLInputElement>) => {
     setStep(Math.round(Number(event.target.value)));
@@ -38,6 +39,10 @@ export const App: React.FC = () => {
     event: React.ChangeEvent<HTMLInputElement>,
   ) => {
     setAnimationDuration(Number(event.target.value));
+  };
+
+  const handleToggleInfinite = () => {
+    setInfinite(!infinite);
   };
 
   return (
@@ -68,6 +73,10 @@ export const App: React.FC = () => {
         value={animationDuration}
         onChange={handleAnimationDuration}
       />
+
+      <button type="button" onClick={handleToggleInfinite}>
+        Toogle Infinite
+      </button>
       {/* Przekazanie images jako props do Carousel */}
       <Carousel
         images={images}
@@ -75,6 +84,7 @@ export const App: React.FC = () => {
         itemWidth={itemWidth}
         frameSize={frameSize}
         animationDuration={animationDuration}
+        infinite={infinite}
       />
     </div>
   );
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,39 +1,97 @@
...
+  const handleSetStep = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setStep(Math.round(Number(event.target.value)));
+  };
+
+  const handleSetItemWidth = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setItemWidth(Math.round(Number(event.target.value)));
+  };
+
+  const handleFrameSize = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setFrameSize(Math.round(Number(event.target.value)));
```
zibi-dev: I think one-liner handlers can be placed inside JSX code. Writing all those types is just cumbersome.

src/components/Carousel.tsx
```diff
@@ -1,26 +1,108 @@
...
+type CarouselProps = {
+  images: string[];
+  step: number;
+  itemWidth: number;
+  frameSize: number;
+  animationDuration: number;
+  infinite?: boolean;
+};
+
+const Carousel: React.FC<CarouselProps> = ({
```
zibi-dev: Here is why u probably shouldn't use React.FC https://medium.com/raccoons-group/why-you-probably-shouldnt-use-react-fc-to-type-your-react-components-37ca1243dd13

src/components/Carousel.tsx
```diff
@@ -1,26 +1,108 @@
...
+        onClick={handlePrevClick}
+        className={`Carousel__button Carousel__button--prev ${offset === 0 && !infinite ? 'Carousel__button--disabled' : ''}`}
+      >
+        &#9664;
+      </button>
+      <button
+        data-cy="next"
+        type="button"
+        onClick={handleNextClick}
+        className={`Carousel__button Carousel__button--next ${offset === maxNextTransform && !infinite ? 'Carousel__button--disabled' : ''}`}
```
zibi-dev: In `package.json`, you will notice that the package `classnames` is installed. Take a look at that library; it's super useful for handling conditional classnames.

src/components/Carousel.tsx
```diff
@@ -1,26 +1,108 @@
...
+    >
+      <ul className="Carousel__list" ref={carouselListRef}>
+        {images.map((image, index) => (
+          <li
+            key={index}
+            style={{ width: `${itemWidth}px`, height: `${itemWidth}px` }}
+          >
+            <img
+              src={image}
+              alt={`${index + 1}`}
```
zibi-dev: Give more descriptive alt attribute

src/components/Carousel.tsx
```diff
@@ -1,26 +1,108 @@
...
+              style={{ width: `${itemWidth}px` }}
+              width={itemWidth}
+              height={itemWidth}
```
zibi-dev, line 92: Since the parent` <li>` element already has a set width and height, there’s no need to specify the width and height for the `` tag in pixels. We can simplify the code by using `width: 100%` and `height: 100%` on the` ` element to make it scale according to the size of the parent `<li>`.

src/components/Carousel.tsx
```diff
@@ -1,26 +1,108 @@
...
+  return (
+    <div
+      className="Carousel"
+      style={{ width: `${frameSize * (itemWidth + gap)}px` }}
+    >
+      <ul className="Carousel__list" ref={carouselListRef}>
+        {images.map((image, index) => (
+          <li
+            key={index}
+            style={{ width: `${itemWidth}px`, height: `${itemWidth}px` }}
```
zibi-dev: It also accepts number values without problems

src/components/Carousel.tsx
```diff
@@ -1,26 +1,108 @@
...
+
+      setOffset(0);
+    }
+  };
+
+  const handlePrevClick = () => {
+    if (carouselListRef.current && offset < 0) {
+      const newOffset = Math.min(offset + step * (itemWidth + gap), 0);
+
+      carouselListRef.current.style.transition = `transform ${animationDuration}ms ease`;
```
zibi-dev: You don't need to apply the animation duration directly to the carousel. You can declare it in the stylesheet and then use the `transitionDuration `property in the style tag within the component.

src/components/Carousel.tsx
```diff
@@ -1,26 +1,108 @@
...
+      carouselListRef.current.style.transform = `translateX(${newOffset}px)`;
```
zibi-dev: This can also be declared in one place within the style attribute in the component. Then you don't need ref for `carouselList`

Zibi95 (changes requested):
Hey, there are a few things to do. I've left comments on what can be improved.

przwojwwp:
without:
            
            
dont pass the tests

src/components/Carousel.tsx
```diff
@@ -1,26 +1,117 @@
...
+    Carousel__button: true,
+    'Carousel__button--prev': true,
+    'Carousel__button--disabled': offset === 0 && !infinite,
+  });
+
+  const nextButtonClasses = cn({
+    Carousel__button: true,
+    'Carousel__button--next': true,
+    'Carousel__button--disabled': offset === maxNextTransform && !infinite,
+  });
```
zibi-dev, line 65: This can be also used like that
```suggestion
    const nextButtonClasses = cn('Carousel__button--next Carousel__button', {
      'Carousel__button--disabled': offset === maxNextTransform && !infinite,
    });
```

Zibi95 (approved):
LGTM!

PR #797, opened 2024-10-07
Solution

Valeraom:
[DEMO LINK](https://Valeraom.github.io/react_carousel/)


BudnikOleksii (approved):
Good work 👍 
But it would be better to add some validation
image

PR #818, opened 2024-12-07
Solutions

OkMoroz:
- [DEMO LINK](https://OkMoroz.github.io/react_carousel/)


BudnikOleksii (approved):
LGTM 🔥 
Just a few suggestions:
- It would be great to add some validations because now a user can set any values
image
- Also, when it's infinite and I click a button fast from the last slide it couldn't get back to the first slide so it stuck a bit

But, great work and code looks good as well

PR #840, opened 2025-01-24
Task solution

Dominik-Dudek-code:
- [DEMO LINK](https://Dominik-Dudek-code.github.io/react_carousel/)


Dominik-Dudek-code:
I have completed the task, except for the additional requirement — infinity. I have a question regarding it: how should the display of images behave in this state? Should it continuously rotate without interruption until the checkbox is unchecked, achieving this using an interval? Or should it involve disabling the "disable" option on the buttons? I'm not entirely sure about the requirements of this point.

natalia-klonowska (approved):
for the infinite requirement the carousel should loop back to the first set of images when the user reaches the end and clicks 'next' allowing continuous navigation through the carousel. the same should happen with the 'previous' button at the start. so both buttons should never be disabled

PR #864, opened 2025-03-12
solution

KateShepel:
[DEMO LINK](https://KateShepel.github.io/react_carousel/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -10,3 +10,13 @@ form {
   flex-direction: column;
   gap: 10px;
 }
+
+label {
+  display: flex;
+  gap: 15px;
+}
+
+input {
+  padding-inline: 5px;
+  width: 50px;
+}
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,12 @@
-// style go here
+h1 {
+  display: block;
+  margin-block: 10px;
+  text-align: center;
+}
+
+form {
+  margin-left: 20px;
+  display: flex;
+  flex-direction: column;
+  gap: 10px;
+}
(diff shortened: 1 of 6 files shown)


src/components/Carousel.scss
```diff
@@ -1 +1,28 @@
-// style go here
+* {
+  margin: 0;
+  padding: 0;
+  text-decoration: none;
+  list-style: none;
+}
+
+.container {
```
alexander-ignatow: This probably should be a container inside `Carousel`

src/components/Carousel.scss
```diff
@@ -1 +1,28 @@
-// style go here
+* {
+  margin: 0;
+  padding: 0;
+  text-decoration: none;
+  list-style: none;
+}
```
alexander-ignatow: You should move this to normalize.scss or reset.scss file

src/components/Carousel.scss
```diff
@@ -1 +1,28 @@
...
+  margin-inline: auto;
+}
+
+.Carousel {
+  &__list {
+    display: flex;
+  }
+}
+
+.button {
```
alexander-ignatow: This should also be an element

src/App.scss
alexander-ignatow, line 1: This looks more like `normalize.scss`

alexander-ignatow (changes requested):
Looks good, but some polish is needed

It was not the task's primary focus, but please, don't leave any inputs like so in the future:
image

And think about validations:
image

alexander-ignatow (approved):
👍

PR #885, opened 2025-04-23
feat: add task solution

s1rserg:
[DEMO LINK](https://s1rserg.github.io/react_carousel/)


s1rserg:
@mateacademy-ai-mentor 
1. I changed value to checked
2. if it's infinite and there are no images to show it resets position to 0
3. It is working ok

PR #977, opened 2025-10-06
_{create carousel with images}_

S-Khraban:
[DEMO LINK](https://S-Khraban.github.io/react_carousel/)


src/components/Carousel.tsx
```diff
@@ -1,26 +1,160 @@
...
+        onClick={goPrev}
+        disabled={!canPrev}
+        aria-label="Previous"
+        className={`Carousel__btn Carousel__btn--prev${!canPrev ? ' is-disabled' : ''}`}
+        style={{
+          position: 'absolute',
+          top: '50%',
+          left: -44,
+          transform: 'translateY(-50%)',
+        }}
```
781calmstate: those styles are not dynamic, why can't you implement them with the class?

src/components/Carousel.tsx
```diff
@@ -1,26 +1,160 @@
...
+                <img
+                  src={src}
+                  alt={`Slide ${i + 1}`}
+                  width={itemWidth}
+                  height={itemWidth}
+                  style={{
+                    width: '100%',
+                    height: itemWidth,
+                    objectFit: 'cover',
+                    display: 'block',
```
781calmstate: non-dynamic styles should be implemented with classes

src/components/Carousel.tsx
```diff
@@ -1,26 +1,160 @@
...
+        style={{ width: viewportWidth, overflow: 'hidden' }}
+      >
+        <ul
+          className="Carousel__list"
+          style={{
+            display: 'flex',
+            gap: 0,
+            margin: 0,
+            padding: 0,
+            listStyle: 'none',
```
781calmstate: non-dynamic styles should be implemented with classes

src/components/Carousel.tsx
```diff
@@ -1,26 +1,160 @@
...
+  const [index, setIndex] = useState(0);
+
+  const canPrev = infinite ? images.length > frameSize : index > 0;
+  const canNext = infinite ? images.length > frameSize : index < maxStart;
+
+  const viewportWidth = frameSize * itemWidth;
+
+  const translateX = -index * itemWidth;
+
+  const goNext = () => {
```
781calmstate: the function naming isn't perfect, the handlers should have prefix `handle-` before the name

src/components/Carousel.tsx
```diff
@@ -1,26 +1,160 @@
...
+
+      return;
+    }
+
+    if (index < maxStart) {
+      setIndex(Math.min(index + safeStep, maxStart));
+    }
+  };
+
+  const goPrev = () => {
```
781calmstate: same as above

src/components/Carousel.tsx
```diff
@@ -1,26 +1,160 @@
...
+  const translateX = -index * itemWidth;
+
+  const goNext = () => {
+    if (!images.length) {
+      return;
+    }
+
+    if (infinite) {
+      if (index >= maxStart) {
+        setIndex(0);
```
781calmstate: again magic number, I don't understand what this `0` represents, why are you setting zero?

if it was some const named for example `INITIAL_POSITION` or `INITIAL_INDEX` that would be more clear + you could change this value just by changing const value instead of searching for every 0 you are using

src/components/Carousel.tsx
```diff
@@ -1,26 +1,160 @@
...
+      } else {
+        setIndex(Math.max(index - safeStep, 0));
+      }
+
+      return;
+    }
+
+    if (index > 0) {
+      setIndex(Math.max(index - safeStep, 0));
+    }
```
781calmstate: a lot of those unclear `zeros`

src/components/Carousel.tsx
```diff
@@ -1,26 +1,160 @@
...
+            listStyle: 'none',
+            transform: `translateX(${translateX}px)`,
+            transition: `transform ${animationDuration}ms`,
+            willChange: 'transform',
+          }}
+        >
+          {images.map((src, i) => {
+            const isVisible = i >= index && i < index + frameSize;
+
+            return (
```
781calmstate: this could be moved into separated component

src/App.tsx
```diff
@@ -20,17 +27,95 @@ class App extends React.Component<{}, State> {
...
       './img/10.png',
     ],
+    itemWidth: 130,
+    frameSize: 3,
+    step: 3,
+    animationDuration: 1000,
+    infinite: false,
   };
 
+  handleNumber<K extends NumKeys>(key: K) {
```
781calmstate: `key: NumKeys` would mean that your `key` can be one of specified `NumKeys`, why do you need generic here?

src/App.tsx
```diff
@@ -20,17 +27,95 @@ class App extends React.Component<{}, State> {
...
+    return (e: React.ChangeEvent<HTMLInputElement>) => {
+      const value = Number(e.target.value);
+      const parsed = Number.isNaN(value) ? 0 : value;
+
+      this.setState({ [key]: parsed } as Pick<State, K>);
+    };
+  }
+
+  handleBoolean(key: 'infinite') {
```
781calmstate: if it can only be the `infinite` what is the sense of passing/accepting this argument?

src/App.tsx
```diff
@@ -20,17 +27,95 @@ class App extends React.Component<{}, State> {
...
 
     return (
       <div className="App">
-        {/* eslint-disable-next-line */}
-        <h1>Carousel with {images.length} images</h1>
+        <h1 data-cy="title">Carousel</h1>
+
+        <div
+          className="controls"
+          style={{ display: 'grid', gap: 12, maxWidth: 420, marginBottom: 24 }}
```
781calmstate: non-dynamic styles should be implemented with classes

781calmstate (changes requested):

src/components/Carousel.tsx
```diff
@@ -1,26 +1,143 @@
...
+type Props = {
+  images: string[];
+  itemWidth?: number;
+  frameSize?: number;
+  step?: number;
+  animationDuration?: number;
+  infinite?: boolean;
+};
+
+const FIRST = 0;
```
781calmstate: first what? :)

src/components/Carousel.tsx
```diff
@@ -1,26 +1,143 @@
...
+      <div className="Carousel__viewport" style={{ width: viewportWidth }}>
+        <ul
+          className="Carousel__list"
+          style={{
+            transform: `translateX(${translateX}px)`,
+            transition: `transform ${animationDuration}ms`,
+          }}
+        >
+          {images.map((src, i) => {
+            const visible = i >= index && i < index + frameSize;
```
781calmstate: i have a feeling that this can be computed inside of the component

src/components/Carousel.tsx
```diff
@@ -1,26 +1,143 @@
...
+const Carousel: React.FC<Props> = ({
+  images,
+  itemWidth = 130,
+  frameSize = 3,
+  step = 3,
+  animationDuration = 1000,
+  infinite = false,
+}) => {
+  const maxStart = Math.max(FIRST, images.length - frameSize);
+  const safeStep = Math.max(1, Math.min(step, Math.max(1, images.length)));
```
781calmstate: still some magic `1`

781calmstate (changes requested):

781calmstate:
You should pay more attention to commits' namings. You can read about best practices [here](https://www.conventionalcommits.org/en/v1.0.0/).

Avoid using those weird symols in commit names (`_`, `{`, `}`)

src/App.tsx
```diff
@@ -20,17 +27,90 @@ class App extends React.Component<{}, State> {
...
+    frameSize: 3,
+    step: 3,
+    animationDuration: 1000,
+    infinite: false,
+  };
+
+  handleNumber(key: NumKey) {
+    return (e: React.ChangeEvent<HTMLInputElement>) => {
+      const value = Number(e.target.value);
+      const parsed = Number.isNaN(value) ? 0 : value;
```
mkyloa, line 40: if you want a shorter form of this
```suggestion
      const value = Number(e.target.value) || 0;
```
but your code looks more descriptive

src/components/Carousel.scss
```diff
@@ -1 +1,78 @@
...
+
+.controls input[type='number'],
+.controls input[type='checkbox'] {
+  justify-self: start;
+}
+
+.controls input[type='number'] {
+  width: 70px;
+  text-align: center;
+}
```
mkyloa, line 78: i would rather set a class for each of these elements
you don't want to complicate the selector if it is not 100% necessary

mkyloa (approved):
awesome job!

PR #978, opened 2025-10-13
solution

ghost:


volodymyr-soltys97 (changes requested):
Hi, here is no demo link 
image

PR #987, opened 2025-11-20
add task solution 20.11.25

Alya-ya:
[DEMO LINK](https://Alya-ya.github.io/react_carousel/)


etojeDenys (changes requested):
almost done

there should be an option on the demo link to make the carousel infinite

etojeDenys (approved):
good job

